PR #763, opened 2015-02-23
Add csv_scenarios:load rake task

twilkinson573:
* Creates a new scenario for each csv file in `config/csv_scenarios`
* Run with `rake csv_scenarios:load`
* Area code currently hardcoded


antw:
I'm not so keen on having to commit the CSV files to the repo; even after deletion Git will retain a copy.

@jorisberkhout Can you clarify why this is needed? If they're scenarios to be kept long-term, we could instead add them as presets. If you just want to add them as a one-off, I think we'd be better adding a CSV import option to the ETEngine [new scenario form](http://beta.et-engine.com/data/latest/scenarios/new).

jorisberkhout:
Good point, @antw. I got a request from somebody who likes to create multiple scenarios at once (for now 12) without having to set all the sliders by hand. I discussed with @dennisschoenmakers and offered this person an Excel sheet where he could fill out the settings of all relevant sliders for all his scenarios at once. From this sheet, I created the 12 CSV files in the pull request (@twilkinson573 modified them slightly).

I had never seen the new scenario form :blush:, but that is exactly what where looking for. I think it is best if @twilkinson573 and myself create the YML files from the CSV files and use the new scenario from.

I suggest you close this pull request. Thanks for the suggestion!

PR #773, opened 2015-06-09
Saved the P2H node position

ChaelKruip:
New nodes P2H and P2G have been given positions in the graph layout.


jorisberkhout:
:+1:

PR #774, opened 2015-06-23
Scaled scenarios

grdw:
For Etmoses to generate a list of all scaled scenarios for the select box on the import page.

* Doesn't take into account the scaled scenario's from a specific user.
* Just gives all scenarios with a set of scaling attributes
* @antw Question: some of the scaled scenarios can contain private information. Is that automatically held into account or should the query be adapted?

### DIFF
--- a/app/controllers/api/v3/scenarios_controller.rb
+++ b/app/controllers/api/v3/scenarios_controller.rb
@@ -31,6 +31,16 @@ module Api
         render json: @scenarios
       end
 
+      def scaled
+        scenarios = Scenario.joins(:scaler)
+
+        @scenarios = scenarios.map do |scenario|
+          ScenarioPresenter.new(self, scenario, params)
+        end
+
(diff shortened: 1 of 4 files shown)


antw:
> @antw Question: some of the scaled scenarios can contain private information.

I don't *think* anything private is revealed through the new action – at least nothing that isn't available through the other actions on `ScenariosController`. I think it's okay. :+1:

ChaelKruip:
@grdw can you assign someone to merge this?

antw:
I think this went unmerged when we decided to take the integration between the ETM and Moses in a different direction: instead of a drop-down with all scaled scenarios in Moses, we click "Open as local energy solution" in the ETM.

This can probably be closed, unless such a drop-down is going to be added to Moses?

ChaelKruip:
> This can probably be closed, unless such a drop-down is going to be added to Moses?

Closing!

PR #793, opened 2015-10-28
New converter position of the added regasification_bio_lng converter

RobTerwel:
And that completes LNG in the ETM for now.


jorisberkhout:
:+1:

PR #795, opened 2015-11-13
FCE for LNG

RobTerwel:
Added LNG to FCE Carriers on ETEngine, included FCE attributes and updated existing attributes.

Note: it was necessary to include the LNG carrier in `fce.rb` to make my local ETM pick it up (otherwise it uses the 'regular' lng.carrier attributes).


jorisberkhout:
:+1:

PR #818, opened 2016-01-28
Adding method to extract the last update for a certain folder/file in git

grdw:
Part of: https://github.com/quintel/etmodel/pull/2002

### DIFF
--- a/app/models/etsource/base.rb
+++ b/app/models/etsource/base.rb
@@ -116,6 +116,13 @@ module Etsource
       nil
     end
 
+    def last_updated_at(folder)
+      if latest_import_sha = get_latest_import_sha
+        git.gcommit(latest_import_sha)
+          .log(1).object(folder).map(&:date).first
+      end
+    end
+
     private
 
     def import_sha_file
--- a/app/models/etsource/loader.rb
+++ b/app/models/etsource/loader.rb
@@ -33,7 +33,7 @@ module Etsource
     #   be independent to reduce bugs.
     def graph(country = nil)
       instrument("etsource.loader: graph") do
-        graph = DeepClone.clone optimized_graph 
+        graph = DeepClone.clone optimized_graph
         graph.dataset = dataset(country) if country
         graph
       end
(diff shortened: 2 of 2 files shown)


grdw:
There's one problem with this code:

````ruby
git.log(1).object(folder).map(&:date).first
````

This sometimes returns the wrong commit date. Now this is in cases when somebody rebased and did some log-rewriting - so I'm not sure wether the github log and the git log are in sync in that case?

**Comment by antw** [comment]
Git has (at least) two commit dates: an `AuthorDate` which will record the time the commit was *authored* (basically, the time of the original commit), and a `CommitDate` which is the time of the commit, cherry-pick, rebase, etc. In most cases, the two will be the same. The Git library is probably returning the former; perhaps there is a method to get the `CommitDate`?

You can see it yourself in the terminal by running `git log --pretty=fuller`.

```
Author:     Anthony Williams <hi@antw.io>
AuthorDate: Wed Jan 27 20:51:27 2016 +0000
Commit:     Anthony Williams <hi@antw.io>
CommitDate: Thu Jan 28 00:11:25 2016 +0000

    Add P2G to Merit flexibility technologies
```

**Comment by grdw** [comment]
*slaps head*

I forgot to do a `git pull` on etsource.. ..thanks for explaining the difference but it returns the correct date. :sweat_smile:

File: app/models/etsource/base.rb [hunk 1]
```diff
@@ -116,6 +116,12 @@ def get_latest_import_sha
       nil
     end
 
+    def last_updated_at(folder)
+      unless Rails.env.test?
+        git.log(1).object(folder).map(&:date).first
```
antw: I think this might not be quite correct: production and staging environments might be running a different ETSource version than is checked out using Git.

`config/config.yml` files have two options: `etsource_working_copy` and `etsource_export`. In development, these both point to the same directory. This means our local ETEngines use ETSource directly.

In staging and production, `etsource_export` has a different path. When someone checks out an ETS version on the [ETSource page](http://beta.et-engine.com/etsource/commits), the contents of that commit are copied into a different directory (`etsource_export`). ETEngine reads data from *that* directory, not the Git clone (`etsource_working_copy`). The live paths are something like:

* `etsource_export`: /var/www/etengine/current/tmp/etsource
* `etsource_working_directory`: /var/www/etsource

It's possible therefore that ETEngine may be using an older ETSource commit:

screen shot 2016-01-28 at 11 15 07

I think you could work around this by supplying the current in-use Git revision:

```ruby
git.gcommit(get_latest_import_sha)
  .log(1).object(path).map(&:date).first
```

grdw: Thanks for your comment! I did know of the import commits functionality but didn't know the technical details :smile:

PR #827, opened 2016-02-12
Added new compressor converter and renamed others

RobTerwel:
see https://github.com/quintel/etdataset/pull/676 and https://github.com/quintel/etsource/pull/1081


jorisberkhout:
:+1:

PR #841, opened 2016-05-19
Make it possible to select custom attributes from et-engine for a converter

grdw:


app/models/api/v3/converter_stats_presenter.rb
```diff
@@ -18,9 +18,10 @@ class ConverterStatsPresenter
 
       attr_reader :key
 
-      def initialize(key, gql)
-        @key = key
-        @gql = gql
+      def initialize(key, gql, graph_attributes)
+        @key              = key
+        @gql              = gql
+        @graph_attributes = graph_attributes || ATTRIBUTES
```
antw: I don't see much point in `|| ATTRIBUTES` any more; the request should supply a list of all attributes wanted. If something like...

```json
{
    "keys": {
        "transport_car_using_electricity": []
    }
}
```

... is requested, just return no attributes; apparently the sender of the request doesn't want anything!

We could actually fail such a request with `400 Bad Request` or `422 Unprocessable Entity`, since it's probably an mistake to ask for no attributes.

app/models/api/v3/converter_stats_presenter.rb
```diff
@@ -31,7 +32,7 @@ def initialize(key, gql)
       end
 
       def as_json(*)
-        ATTRIBUTES.each_with_object({}) do |attr_key, hash|
+        @graph_attributes.each_with_object({}) do |attr_key, hash|
```
antw: These attributes are coming straight from the request without any validation, and then being executed by `public_send`! Luckily not much damage can be done with `public_send` and no arguments, but it's still risky.

Check that each attribute is permitted before executing:

```ruby
def as_json(*)
  unless @graph_attributes.all? { |key| PERMITTED_ATTRIBUTES.include?(key) }
    fail "Not gonna happen!"
  end

  @graph_attributes.each_with_object({}) do |attr_key, hash|
    hash[attr_key] = {
      present: @present.public_send(attr_key),
      future:  @future.public_send(attr_key)
    }
  end
end
```

app/controllers/api/v3/converters_controller.rb
```diff
@@ -24,13 +24,18 @@ def show
       # converter keys, the request is to be sent as POST with a JSON payload
       # with the following schema:
       #
-      # { "keys": [ "key1", "key2", "...", "keyN" ] }
+      # { "keys": [
+      #   { "key1": [ .. ],
+      #   { "key2": [ .. ] },
+      #   "...",
+      #   { "keyN": [ .. ] }
+      # ] }
```
antw: `"keys"` is expected to be a hash/object, not an array:

```
{
  "keys": {
    "key1": [ .. ],
    "key2": [ .. ],
    "...",
    "keyN": [ .. ]
  }
}
```

antw:
I resolved the conflicts and merged manually. Thanks @grdw!

PR #849, opened 2016-06-17
Fixes the Infinity CSV bug

grdw:
Fixes: https://github.com/quintel/etmodel/issues/2076

### DIFF
--- a/app/models/qernel/plugins/merit/producer_adapter.rb
+++ b/app/models/qernel/plugins/merit/producer_adapter.rb
@@ -36,14 +36,7 @@ module Qernel::Plugins
       def producer_attributes
         attrs = super
 
-        marginal_cost = @converter.variable_costs_per(:mwh_electricity)
-
-        attrs[:marginal_costs] =
-          if marginal_cost && marginal_cost.nan?
-            Float::INFINITY
-          else
-            marginal_cost
-          end
+        attrs[:marginal_costs] = @converter.marginal_costs
 
         attrs[:output_capacity_per_unit] =
           @converter.electricity_output_conversion * @converter.input_capacity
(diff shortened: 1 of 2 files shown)


antw:
Coincidentally this was also fixed by the [recent change to merit order pricing](https://github.com/quintel/etengine/commit/da7fec09de28989795141573dfbb90515a101c4d), however I'm still going to merge this. I like the change to use `Converter#marginal_costs`; I can't find any situation in which it differs from `variable_costs_per(:mwh_electricity)`. I tried changing efficiencies and fuel mixes and the two were always identical.

I'm not sure why `variable_costs_per` was used instead; I can only assume that `marginal_costs` didn't exist when the merit order was first implemented, and it never got changed ([17 Sep 2012](https://github.com/quintel/etengine/commit/0eb46888c4dce923b9e9288d208521d63e16ecd7) vs [12 Dec 2012](https://github.com/quintel/etengine/commit/b561e9cbfcb505ed0e9ca3c1d549752f3344ed61)).

ChaelKruip:
> I can't find any situation in which it differs from variable_costs_per(:mwh_electricity). I tried changing efficiencies and fuel mixes and the two were always identical.

They are identical, except for when the production of a plant equals zero. In that case a division by zero results in Infinity for the `variable_costs_per(:mwh_electricity)` method while the `marginal_costs` method is robust against that.

> I can only assume that marginal_costs didn't exist when the merit order was first implemented,

Exactly!

antw:
> In that case a division by zero results in Infinity

Division by zero was only half the problem. It was division *of* zero (`variable_costs`) *by* zero (`typical_electricity_output`) which creates a `NaN`. That's a problem for Merit since it needs to sort dispatchables by their cost and `NaN` isn't a number. Therefore we'd convert `NaN` to Infinity, which served us quite well until the need to export the prices as CSV files (or plot them in a chart).

PR #874, opened 2016-11-03
update installation instructions and merge the two sections

ploh:


README.markdown
```diff
@@ -22,29 +22,50 @@ The ETE is released under the [MIT License](LICENSE.txt).
...
+   ```
+   Then make any changes -- particularly to the database configuration -- as you see fit.
+   * Probably set "standalone" to `true` in "config/config.yml"
 
-4. Clone a copy of [ETSource][etsource] –– which contains the data for each
-   region. Edit "config/config.yml" and enter the ETSource directory into the
-   "etsource_export" and "etsource_working_copy" options.
+1. Clone a copy of [ETSource][etsource] –– which contains the data for each
+   region:
+   1. `cd ..; git clone git@github.com:quintel/etsource.git`
```
grdw: This documentation probably belongs in ETSource and not in ETEngine.

antw: A local copy of ETSource is needed to run ETEngine, so I think that line makes sense.

There isn't any setup required to use ETSource with ETEngine other than cloning the repo. If you want to make *changes* to ETSource then that's another matter; you'd want to run `bundle install` so that the tests can be performed, however I don't think that's a necessary part of setting up ETEngine.

ploh:
@grdw @antw Do you agree with this now (I have not changed it any more)?

PR #875, opened 2016-11-03
explicitly set the ruby version to 2.1.2

ploh:
currently, the bundled gems can be installed also with ruby 2.3...
but because of a bug in ruby_deep_clone there will be runtime errors


antw:
Thanks @ploh!

You can ignore the Semaphore error, it's configured to use 2.1.0 (!!!). I can change the Ruby version used so it's not a problem.

Can we maybe aim for Ruby 2.1.9 instead? 2.1.2 [is especially old](https://www.ruby-lang.org/en/news/2014/05/09/ruby-2-1-2-is-released/) and there have been many security and bug fixes between 2.1.2 and 2.1.9. (Ruby 2.1.10 is a test version not intended for production use, so 2.1.9 remains the latest 2.1.x).

ploh:
changed to 2.1.9, all specs still running

antw:
I gave Semaphore CI a push in the right direction and it no longer complains. Thanks for this! :+1:

PR #876, opened 2016-11-03
Bump mysql2 gem version to 0.3.21

ploh:
Rationale: Incompatibility of old gem version with newer mysql version


antw:
Thanks. I expect I should update my local version.

> mysql  Ver 14.14 Distrib **5.6.10, for osx10.8 (x86_64)**

PR #878, opened 2016-11-04
Use rbenv specification for capistrano version specification

ploh:


ploh:
Maybe we do it like this to be sure it always works

ploh:
Unfortunately, does not work with semaphore

PR #886, opened 2016-12-13
Flag derived datasets in areas.json

ploh:


app/models/etsource/loader.rb
```diff
@@ -46,8 +46,11 @@ def presets
 
     def area_attributes(area_code)
       @area_attributes[area_code] ||= begin
-        area_attr = Atlas::Dataset.find(area_code).to_hash
+        area = Atlas::Dataset.find(area_code)
+        area_attr = area.to_hash
+        area_attr[:is_derived] = true if area.is_a?(Atlas::Dataset::DerivedDataset)
```
grdw: This can be roughly translated to:

```ruby
area_attr[:is_derived] = area.is_a?(Atlas::Dataset::DerivedDataset)
```

ploh: First I thought you were right and I made a braindead mistake here, but then I remembered my reason for this: I only want to have this key in the hash if it is true, but not if it is false

grdw: > I only want to have this key in the hash if it is true, but not if it is false

That's because of [this line](https://github.com/quintel/etmodel/pull/2313/files#diff-f8737ffcf0d1a40db31e52e65222b993R47) I guess? But I think it's also perfectly acceptable to have `false` there for `is_derived?` I guess. Unless it's used in other places?

ploh: It is not (yet) used in other places. I am fine if you change it because the code looks ugly now :smile:

antw: I agree with @grdw. Also, I don't think it's right to have the `is_` prefix; such prefixes are normally omitted in Ruby, especially in predicate methods.

For example:

* `thing.name`, not `thing.get_name`
* `thing.name=`, not `thing.set_name`
* `thing.empty?` not `thing.is_blank?`
* **`thing.derived?` not `thing.is_derived?`**

Otherwise, once the tests pass I think this can be merged.

ploh: > Also, I don't think it's right to have the is_ prefix; such prefixes are normally omitted in Ruby, especially in predicate methods.

Yes, I know - and that is one of the many things I love about Ruby.

I prepended `is_` here, because I am not defining a Ruby method. Instead I am filling a hash which is gonna be transferred as json. And in this we probably do not want to have a "?". There also is precedence for this: The countless `has_...` attributes of a dataset which could also be question-marked in Ruby.

But I guess you may as well argue the other way around and say it is fine to just leave out the prefix `is_` without adding the postfix "?". There is precedence for this as well, with the `useable` (virtual) attribute.
So I will remove the prefix!

PR #890, opened 2016-12-22
Local datasets exporter inheritance

ploh:
Use Atlas::FullExporter (because Atlas::Exporter is an abstract base now)

Ref quintel/atlas@88a590aab1e0ec5cc
Ref quintel/atlas#66


ploh:
@grdw This needs quintel/atlas#66 to be merged first. And then we should update the used Atlas version in the ETEngine Gemfile.

PR #894, opened 2017-01-09
Initializer inputs

grdw:
This is the most basic implementation of the initializer inputs. 

Related to: https://github.com/quintel/atlas/pull/81


ploh:
@grdw I pushed my suggestions for this. Please review, further adapt if necessary and merge. Thx.

app/models/gql/gql.rb
```diff
@@ -1,337 +1,350 @@
...
+        "#<#{ self.class.name } (unknown scenario)>"
+      end
     end
-  end
 
-end
+    private
 
+    def apply_initializer_inputs(graph = nil)
+      unless graph
```
grdw: This is a bit of a werid solution to be honest, why cramp everyting in a single method?

ploh: Yeah, it seemed like a cool idea earlier, but now at second glance it looks weird!

app/models/qernel/graph.rb
```diff
@@ -402,6 +402,17 @@ def graph_query
     @graph_query ||= GraphApi.new(self)
   end
 
+  def initializer_inputs
+    if init = area.init
+      init.
+        map { |key, val| [Input.fetch(key), val] }.
+        sort_by { |input, _| input.priority }.
+        reverse
```
grdw: I'm a bit of a fan of splitting these things up in several methods like:

````ruby
def initializer_inputs
  decorated_inputs.sort_by { |input, _| input.priority }.reverse
end

private

def decorated_inputs
  (area.init || {}).map do |key, val|
    [Input.fetch(key), val]
  end
end
````

... or something like that.

PR #898, opened 2017-01-12
Local datasets scaling of input sliders

grdw:
This is done as is!

---

~~First merge  https://github.com/quintel/etengine/pull/894~~

### DIFF
--- a/app/models/scenario_scaling.rb
+++ b/app/models/scenario_scaling.rb
@@ -24,6 +24,14 @@ class ScenarioScaling < ActiveRecord::Base
     new(scaling.attributes)
   end
 
+  def self.from_scenario(scenario)
+    if scenario.area[:derived]
+      from_atlas_scaling(scenario.area[:scaling])
+    else
+      scenario.scaler
+    end
+  end
+
   # Public: Determines if the given input should have it's values scaled.
   #
   # Returns true or false.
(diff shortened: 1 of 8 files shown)


app/models/scenario_scaling.rb
```diff
@@ -24,6 +24,14 @@ def self.from_atlas_scaling(scaling)
     new(scaling.attributes)
   end
 
+  def self.from_scenario(scenario)
+    if scenario.area['derived']
```
antw: The area is a `HashWithIndifferentAccess`, so you can use Symbols for keys instead of allocating extra strings (same applies when Scenario reads `scaling`).

spec/models/input_spec.rb
```diff
@@ -25,7 +25,14 @@
         input.min_value_for(scenario).should eql(2)
       end
 
-      it 'should return scaled values when the scenario is scaled' do
+      it 'should return scaled values when the scenario is scaled by area code' do
```
antw: Since there are separate tests for "scaled by scaler" and "scaled by area code", does it make sense to split these into two separate context blocks so that common setup code can be shared between examples?

```ruby
# ...

context 'with an unscaled scenario' do
  it 'computes the min value'
  it 'computes the max value'
  it 'computes the start value with a Gql::Gql'
  it 'computes the start value with a Scenario'
  # etc ...
end

context 'with a scenario scaled to 1000 households' do
  before do
    scenario.create_scaler!(
      area_attribute: 'number_of_residences',
      value: 10000
    )
  end

  it 'scales down the min value'
  it 'scales down the max value'
  it 'scales down the start value with a Gql::Gql'
  it 'computes the start value with a Scenario'
  # etc ...
end

context 'with a derived dataset scaled to 2000 households' do
  before { scenario.update_attribute(...) }

  it 'scales down the min value'
  it 'scales down the max value'
  it 'scales down the start value'
  # etc ...
end
```

That might also provide a good excuse to switch this file from the old `should`-style assertions to `expect`?

grdw:
@antw I managed to fix your points 👍 

Would it be okay to merge?

PR #900, opened 2017-01-17
Make it possible to scale a derived dataset

grdw:
If the area of the scenario is derived.


app/controllers/api/v3/scenarios_controller.rb
```diff
@@ -88,7 +88,7 @@ def create
 
         @scenario = Scenario.new
 
-        if scaler_attributes && ! attrs[:descale]
+        if scaler_attributes && ! attrs[:descale] && ! Area.derived?(attrs.fetch(:area_code))
```
antw: * Any chance of a test for this? [There are tests for other two conditions](https://github.com/quintel/etengine/blob/470256764853fa338a13543e272de324bd46610b/spec/requests/api/v3/create_scenario_spec.rb#L217-L320) (scaler attributes being provided and/or the `descale` option).

* It seems like this would make impossible to (further) scale a derived region into something smaller. Would this not be useful to have in the event that you want to scale a derived scenario for use in Moses? For example, perhaps to scale Ameland into a sub-region representing only a couple of streets?

  I guess the problem is that this means having two sets of scaling settings (one for the region, one for the scenario), and that might be a pain to manage. I guess the expectation is that one day we'll do away with the old method of scaling entirely, but that seems quite far off... @ChaelKruip does it seem okay to you to *not* be able to (re-)scale derived scenarios? I can imagine it being annoying to some users.

app/models/area.rb
```diff
@@ -3,4 +3,9 @@ def self.get(area_code)
     attributes = Etsource::Loader.instance.area_attributes(area_code)
     attributes.with_indifferent_access
   end
+
+  def self.derived?(area_code)
+    Atlas::Dataset.exists?(area_code) &&
+      get(area_code).fetch(:derived)
```
antw: Looks like this is short enough to put on one line without exceeding 80 characters?

app/models/scenario.rb
```diff
@@ -119,7 +119,11 @@ def area
   end
 
   def scaled?
-    scaler.present? || area[:derived]
+    scaler.present? || derived?
+  end
+
+  def derived?
```
antw: I get where you're going with this, but it feels wrong to ask the scenario if *it* is derived. The scenario isn't derived; the area is. Preferably we could do `scenario.area.derived?` but that isn't possible on account of `area` being a hash (for reasons that escape me).

Perhaps this method could be called something like `area_derived?`?

It feels a *bit* unnecessary since it is called from only one place. It could instead be abstracted away in `Area`.

```ruby
class Area
  def self.derived?(area)
    area[:derived]
  end

  def self.derived_code?(area_code)
    Atlas::Dataset.exists?(area_code) && derived?(get(area_code))
  end

  # ...
end
```

 I have no strong preference either way.

app/controllers/api/v3/scenarios_controller.rb
grdw: > (scaler attributes being provided and/or the descale option).

Should it be possible to descale a scenario from a derived dataset? What would you even expect? To get a scenario with the `base_dataset` and the inputs scaled up to the size of Holland?

ChaelKruip: > @ChaelKruip does it seem okay to you to not be able to (re-)scale derived scenarios? I can imagine it being annoying to some users.

Hm. I think it should be possible still. People may want to open a scenario for a municipality and scale it to their neighbourhood or street etc. Right @AlexanderWirtz?

antw: > Should it be possible to descale a scenario from a derived dataset? What would we expect? To get a scenario with the base_dataset and the inputs scaled up to the size of Holland? 🤔

If you were to scale a scenario which uses a derived dataset (e.g. to 10 households), then descaling would return it back to the size of the *derived* region (Ameland), not the original NL dataset.

It should not be possible (at this stage) to take a scenario with a derived dataset and convert it to a full-size NL scenario.

As far as ETEngine is concerned, I think it should see the derived dataset and the original dataset as two entirely different things: they're just two different datasets; how Atlas+Refinery created them is not important.

grdw: I think that this code will not correctly work with the idea that a scenario from a derived dataset can scale. Mainly because of [these lines](https://github.com/quintel/etengine/blob/local-datasets-scaling-of-sliders/app/models/scenario_scaling.rb#L27-L33). 

I'll try to spec this though and see if that'll work. But it's good to know that descaling will descale back to the `derived_dataset` area. 👍

grdw:
I managed to solve it through ETModel though instead of doing a bit of a hack inside of ETEngine. See https://github.com/quintel/etmodel/pull/2320. I still need to test the scaling of a derived dataset though.

grdw:
I'm going to close this for now. This needs a proper implementation.

PR #903, opened 2017-01-20
Add the possibility to scale derived scenarios

grdw:
A pull request to scaled derived datasets. This also includes some clean-up to the `Input` class.

Fixes: https://github.com/quintel/etsource/issues/1215


grdw:
This doesn't work yet @antw 😆 but maybe you could take a look why it doesn't. The caching fetches the incorrect values somehow ... 🤔 but I modified the cache a little so it caches the `scaled` values for a Derived dataset. That seems to be working, but scaling an already scaled scenario (scale-ception (being derived or not derived)) is currently not working as expected (it looks as though the cache is not properly removed or it reads old values .... I'm not sure though ).

If you have time; could you take a quick peek into what seems to be the problem.

antw:
@grdw Is there a specific ETSource branch or config I need to run this? I tried with [local-datasets-ameland-mock-dataset](https://github.com/quintel/etsource/tree/local-datasets-ameland-mock-dataset), but I get an error whenever I try to start a scenario in "ameland". I tried creating a new derived dataset in case "ameland" was out-of-date, but that has the same problem.

It appears this is because the "agriculture_chp_engine_biogas" node (a CentralProducer) is missing the a value for `full_load_hours`, which is one of the values [supposedly exported by Atlas](https://github.com/quintel/atlas/blob/c346a75fadc70f6038459b3b03b8b373169fcd90/lib/atlas/node/central_producer.rb#L10) (it is likely this affects all central producers; just that this is the first one used). Indeed, "ameland.pack" does not have a value, but "nl.pack" does:

```ruby
MessagePack.load(File.read('tmp/atlas/ameland.pack'))['nodes']['agriculture_chp_engine_biogas']['full_load_hours']
# => nil

MessagePack.load(File.read('tmp/atlas/nl.pack'))['nodes']['agriculture_chp_engine_biogas']['full_load_hours']
# => 8000.0
# https://github.com/quintel/etsource/blob/master/datasets/nl/central_producers.csv#L38
```

I don't see any sign of the attribute in the graph.yml in ETSource; I think it should be there since it is a dynamic attribute. I suspect that is the source of the problem, although I'm not sure why I have this issue but other's don't. 😕 

```
NoMethodError - undefined method `*' for nil:NilClass:
  app/models/qernel/converter_api/capacity_production.rb:65:in `full_load_seconds'
  app/models/qernel/converter_api/helper_calculations.rb:15:in `block in number_of_units'
  app/models/qernel/dataset_attributes.rb:208:in `fetch'
  app/models/qernel/converter_api/helper_calculations.rb:12:in `number_of_units'
  app/models/qernel/plugins/merit/adapter.rb:47:in `producer_attributes'
  app/models/qernel/plugins/merit/producer_adapter.rb:37:in `producer_attributes'
  app/models/qernel/plugins/merit/always_on_adapter.rb:7:in `producer_attributes'
  app/models/qernel/plugins/merit/adapter.rb:30:in `participant'
  app/models/qernel/plugins/simple_merit_order.rb:72:in `block in setup'
  app/models/qernel/plugins/simple_merit_order.rb:122:in `block in each_adapter'
  app/models/qernel/plugins/simple_merit_order.rb:122:in `each_adapter'
  app/models/qernel/plugins/simple_merit_order.rb:71:in `setup'
  app/models/qernel/plugins/plugin.rb:28:in `block in trigger'
  app/models/qernel/plugins/plugin.rb:26:in `trigger'
  app/models/qernel/lifecycle.rb:62:in `block in with_callback'
  app/models/qernel/lifecycle.rb:61:in `with_callback'
  app/models/qernel/lifecycle.rb:26:in `calculate'
  app/models/qernel/graph.rb:289:in `calculate'
  app/models/gql/gql.rb:274:in `calculate_graphs'
  app/models/gql/gql.rb:214:in `block in prepare'
  lib/instrumentable.rb:5:in `block in instrument'
  activesupport (4.1.5) lib/active_support/notifications.rb:159:in `block in instrument'
  activesupport (4.1.5) lib/active_support/notifications/instrumenter.rb:20:in `instrument'
  activesupport (4.1.5) lib/active_support/notifications.rb:159:in `instrument'
  lib/instrumentable.rb:4:in `instrument'
  app/models/gql/gql.rb:214:in `prepare'
  app/models/scenario.rb:170:in `gql'
  app/controllers/data/base_controller.rb:21:in `initialize_gql'
```

grdw:
I think you should just use `local-datasets` in ETSource. The `local-datasets-ameland-mock-dataset` should be removed from ETSource as a branch because I pushed it as en example for Berend.

antw:
> but I modified the cache a little so it caches the scaled values for a Derived dataset [...] but scaling an already scaled scenario (scale-ception (being derived or not derived)) is currently not working as expected (it looks as though the cache is not properly removed or it reads old values .... I'm not sure though ).
>
> If you have time; could you take a quick peek into what seems to be the problem.

That's your problem. 😉 

Previously, `Input.cache` looked like this:

```ruby
def self.cache(scenario = nil)
  if scenario && scenario.scaler
    scenario.scaler.input_cache
  else
    @_cache ||= Input::Cache.new
  end
end
```

... and now it looks like this:

```ruby
def self.cache(scenario = nil)
  if scenario && scenario.scaler
    @input_cache ||= ScaledInputs.new(Input.cache, scenario.gql)
  else
    @_cache ||= Cache.new
  end
end
```

Previously, a scaled scenario used an input cache which was memoized on the scaler; this meant that the cache was scoped to the scenario. This is necessary since scenarios could be scaled to any value (therefore their values cannot actually be cached). You've moved the memoization into `Input.cache` and in doing so made the cache global: every scaled scenario will use it.

grdw:
@antw Yes, that's the one! Probably related: https://github.com/quintel/etsource/issues/1215

antw:
This looks pretty good to me: scaling and re-scaling derived scenarios seems to work nicely. I do have one small issue:

1. I created a dataset derived from NL with 10000 households as the scaling value.
2. Opened a new scenario using that derived dataset.
3. ETModel is showing unscaled inputs:

  screen shot 2017-01-24 at 13 39 21

  `App.settings.get_scaling()` returns `null`. Is that supposed to return the derived dataset scaling? I suspect this happens because ETModel has a fixed-coded step value of 0.1 (defined in the DB) and there's no scaling present to adjust the step. The values coming out of ETEngine look to be correctly scaled, so **perhaps this is intentional** and you mean to [get rid of the hard-coded ETM step](https://github.com/quintel/etsource/issues/1213#remove-etm-step)?

```json
{
  "min": 0.011262520978486833,
  "max": 0.04505008391394733,
  "default": 0.022525041956973666,
  "step": 0.0001,
  "code": "households_number_of_inhabitants",
  "unit": "#"
}
```

FYI: I scaled the scenario to 1k and then re-scaled to 100 households; that works perfectly:

screen shot 2017-01-24 at 13 39 26
screen shot 2017-01-24 at 13 39 30

grdw:
> App.settings.get_scaling() returns null. Is that supposed to return the derived dataset scaling?

[It should](https://github.com/quintel/etmodel/commit/283b64e16683b499ae1d644488277e912e9e649d#diff-cc6e703301c6ed33065f186cec0d2939R76). I'm wondering why it doesn't though. 🤔

antw:
> It shouldn't. I'm wondering why it doesn't though. 🤔

I think I have a clue:

* If I start a new scenario using the derived dataset from the front-page (the region/end year form) everything works as expected.

* If I start a scenario by cloning an existing one (etmodel.dev/scenarios/:id), the scaling is `null`.

antw:
> I think I have a clue: [...]

Oh. It looks like there's a whole different code path when starting an existing scenario from `scenarios/:id`:

* [From the front-page](https://github.com/quintel/etmodel/blob/2c9b4514e42a793611e38830dc51d9f4a0fbe641/app/controllers/pages_controller.rb#L59-L70).
* [From scenarios/:id](https://github.com/quintel/etmodel/blob/2c9b4514e42a793611e38830dc51d9f4a0fbe641/app/controllers/scenarios_controller.rb#L97-L122)

I should have refactored that when I implemented scaling. 😞  I think we might be best combining those two implementations so that there's only one.

PR #905, opened 2017-01-23
Implement initializer inputs as a loose class

grdw:
~~Based on https://github.com/quintel/atlas/pull/86~~

---

[Relevant line](https://github.com/quintel/etengine/pull/905/files#diff-ae7d1fad20062c85df9878da0d60dab0R406) for `graph.rb`, the rest is all indentation.


app/models/gql/gql.rb
```diff
@@ -335,14 +335,13 @@ def inspect
     private
 
     def apply_initializer_inputs
-      set_initializer_inputs(:present)
-      set_initializer_inputs(:future)
+      %i(present future).each(&method(:set_initializer_inputs))
```
antw: Personally, I think this is less readable than it was before. I also recently learned that `&method(name)` isn't such a good idea since it performs quite poorly compared to the more explicit `each { |thing| name(thing) }`.

```
Calculating -------------------------------------
             &method    388.031k (± 4.2%) i/s -      1.963M in   5.067859s
        each { ... }      1.779M (± 4.2%) i/s -      8.918M in   5.021471s
```

app/models/base_input.rb
```diff
@@ -0,0 +1,39 @@
+class BaseInput
```
antw: How would you feel about this being a module – included in `Input` and `InitializerInput` – rather than a base class? It seems like both classes need to share some implementation (hence a module) but [cannot be substituted](https://en.wikipedia.org/wiki/Liskov_substitution_principle) for one another.

grdw: I don't feel convinced yet that this is a better approach. I tried it, but I saw that there were some conflicts with the `self.<method>` methods. Probably for that to work I have to turn it into a `Concern`?

antw: > Probably for that to work I have to turn it into a `Concern`?

You can do it manually, but `Concern` would make it simpler:

```ruby
# app/models/input/common.rb
class Input
  module Common
    extend ActiveSupport::Concern

    included do
      include InMemoryRecord
      include CommandAttributes
      
      extend  ActiveModel::Naming

      attr_accessor :file_path
    end

    def initialize(attrs = {})
      # ...
    end

    def key=(new_key
      # ...
    end

    # ...

    module ClassMethods
      def load_records
        # ...
      end

      # ...
    end
  end
end

# app/models/input.rb
class Input
  include Common
  # ...
end

# app/models/initializer_input.rb
class InitializerInput
  include Input::Common
  # ...
end
```

app/models/qernel/initializer_inputs.rb
```diff
@@ -0,0 +1,19 @@
+module Qernel
+  class InitializerInputs
+    def self.for_area(area_code)
+      new(area_code).all
+    end
+
+    def initialize(area_code)
+      @area_attributes =
+        Etsource::Loader.instance.area_attributes(area_code)[:init] || {}
```
ploh: `Etsource::Loader#area_attributes()` will directly access the `Atlas::Dataset`, although the dataset is already loaded in `graph.dataset`. Hence, I strongly believe the loading of the init. inputs for a particular derived dataset should remain in graph.rb (or area.rb).

app/models/qernel/initializer_inputs.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    def all
+      Hash[@area_attributes
+        .map { |key, val| [InitializerInput.fetch(key), val] }
+        .sort_by { |input, _| -input.priority }
```
ploh: Please include the `key` as secondary sort parameter

app/models/qernel/initializer_inputs.rb
```diff
@@ -0,0 +1,19 @@
+module Qernel
+  class InitializerInputs
```
ploh: If not inside graph.rb (see below), then at least put this in dataset.rb

app/models/base_input.rb
ploh: > It seems like both classes need to share some implementation (hence a module) but [cannot be substituted](https://en.wikipedia.org/wiki/Liskov_substitution_principle) for one another.

The LSP does not apply here, it is about subtypes, not "sibling-types" :wink:
But I agree that the base class does not feel like a class - e.g. it does not have any data in it. Hence, I am in favour of making this a module. :+1:

app/models/qernel/graph.rb
```diff
@@ -1,450 +1,438 @@
...
-  end
+    def graph_query
+      @graph_query ||= GraphApi.new(self)
+    end
 
-  def decorated_inputs
-    (area.init || {}).map do |input_key, input_value|
-      [Input.fetch(input_key), input_value]
+    def initializer_inputs
+      @initializer_inputs ||= InitializerInputs.for_area(area.area_code)
```
ploh: I am not sure whether this will - unintentionally - survive a cache flush, e.g. an Etsource::Reloader call.

app/models/qernel/initializer_inputs.rb
grdw: I was initially against putting it in `graph.rb` because the file is already over 400 lines of code which is my opinion is too long for any class.  But you make a valid point. Maybe someday, the `graph.rb` needs some clean-up.

PR #907, opened 2017-01-26
Derived datasets: Preparations for migrating scenarios

ploh:
Some preparations to aid in the migration of existing ETE scaled scenarios into new DerivedDataset based scenarios:
- `Etsource::Reloader#stop!` to manually control the reloading of etsource changes directly after creating a new derived dataset
- Make a graph calculation more robust to rounding errors - proved to be a problem in a test
- Disable caching DatasetAttributes#fetch when applying inputs, so that the same input can be applied twice with different values
- Sort inputs by priority and then also by key, so that the sorting is deterministic
- Use new Atlas version:
  - No scaling of values derived from the interconnector_capacity - to be consistent with the ETE scenario scaling
  - Allow to explicitly set a scaling base_value (when creating a derived dataset) different from the area's number_of_residences - to allow the same scaling factor as in legacy ETE scenarios

### DIFF
--- a/app/models/scenario/user_updates.rb
+++ b/app/models/scenario/user_updates.rb
@@ -93,7 +93,7 @@ module Scenario::UserUpdates
   def inputs
     @inputs ||=
       combined_values.map { |key, _| Input.get(key) }.
-      compact.sort_by(&:priority).reverse
+      compact.sort_by { |input| [-input.priority, input.key] }
   end
 
   # Internal: A hash of inputs, and the values to be set on the named graph.
--- a/Gemfile
+++ b/Gemfile
@@ -43,7 +43,7 @@ gem 'rubel',         ref: 'e36554a',   github:  'quintel/rubel'
 gem 'quintel_merit', ref: 'b9a2147',   github:  'quintel/merit'
 gem 'turbine-graph', '>=0.1',          require: 'turbine'
 gem 'refinery',      ref: '58c1138',   github: 'quintel/refinery'
-gem 'atlas',         ref: 'b76cba4',   github: 'quintel/atlas'
+gem 'atlas',         ref: 'c346a75',   github: 'quintel/atlas'
 
 # system gems
 gem 'mysql2',         '~>0.3.11'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -12,8 +12,8 @@ GIT
 
 GIT
   remote: git://github.com/quintel/atlas.git
-  revision: b76cba44f59b9f5a2592a267998c673daf980823
-  ref: b76cba4
+  revision: c346a75fadc70f6038459b3b03b8b373169fcd90
+  ref: c346a75
   specs:
     atlas (1.0.0)
       activemodel (~> 4.0)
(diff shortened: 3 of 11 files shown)


app/models/scenario/user_updates.rb
```diff
@@ -93,7 +93,7 @@ def user_values_as_yaml=(values)
   def inputs
     @inputs ||=
       combined_values.map { |key, _| Input.get(key) }.
-      compact.sort_by(&:priority).reverse
+      compact.sort_by { |input| [-input.priority, input.key] }
```
grdw: I think `compact.sort_by { |input| -input.priority }` will just work aswell.

antw: I think @ploh's aim was that two inputs with identical priorities would always execute in the same order, regardless of the order in which they are specified in `user_values`.

antw:
There are rather a lot of failing tests, but I don't see any relation between the failures and the changes made in this pull request. Are these due to changes in the `local-datasets` branch?

grdw:
> Are these due to changes in the local-datasets branch?

The non `success?` / status-esque tests that are failing:

```ruby
  12) Updating inputs with API v3 when autobalance=false, providing a balanced single value sets the user value
     Failure/Error: expect(scenario.user_values).to include('balanced_one' => 100.0)
       expected {"unrelated_one"=>25.0} to include {"balanced_one"=>100.0}
       Diff:
       @@ -1,2 +1,2 @@
       -[{"balanced_one"=>100.0}]
       +"unrelated_one" => 25.0

     # ./spec/requests/api/v3/update_input_spec.rb:115:in `block (4 levels) in <top (required)>'

  .....

  21) Updating inputs with API v3 when autobalance=true, providing one member of a group sets the user value
     Failure/Error: expect(scenario.user_values).to include('balanced_one' => 10.0)
       expected {"unrelated_one"=>25.0} to include {"balanced_one"=>10.0}
       Diff:
       @@ -1,2 +1,2 @@
       -[{"balanced_one"=>10.0}]
       +"unrelated_one" => 25.0

     # ./spec/requests/api/v3/update_input_spec.rb:278:in `block (4 levels) in <top (required)>'

  22) Updating inputs with API v3 when autobalance=true, providing one member of a group sets the balanced value
     Failure/Error: expect(scenario.balanced_values).to include('balanced_two' => 90.0)
       expected {"unrelated_two"=>75.0} to include {"balanced_two"=>90.0}
       Diff:
       @@ -1,2 +1,2 @@
       -[{"balanced_two"=>90.0}]
       +"unrelated_two" => 75.0

     # ./spec/requests/api/v3/update_input_spec.rb:282:in `block (4 levels) in <top (required)>'
```

ploh:
The test failures are due to the deterministic input sort. I will investigate how that is related

ploh:
> The test failures are due to the deterministic input sort. I will investigate how that is related

Some of the tests simply forgot to set the `priority` of a fixture `Input`. The old sorting code would also have had a problem, unless all the compared priorities are equally unset - which was the case in the tests.

ploh:
@antw @grdw Maybe we should make the `priority` of `Input` and `InitializerInput` a required attribute!?

PR #908, opened 2017-01-30
Local datasets

grdw:
A pull request for all things related to local datasets. This can be merged without the support of ETSource and Atlas as far as I'm aware.


app/models/etsource/inputs.rb
```diff
@@ -1,20 +1,23 @@
 module Etsource
-
   class Inputs
-    def initialize(etsource = Etsource::Base.instance)
-      @etsource = etsource
+    INPUTS = {
```
antw: Sorry for the continued nitpicking  (😞), but it feels a bit "magic" to initialize this with an Atlas class and then have an ETEngine class automatically selected to receive the data. I think it would be clearer if it were explicit:

* Provide both the Atlas and ETEngine classes
  
  ```ruby
  Etsource::Inputs.new(etsource, Atlas::Input, Input).import
  Etsource::Inputs.new(etsource, Atlas::InitializerInput, InitializerInput).import
  ```

* **Or** provide neither, and use a symbol key to identify which you are importing (still somewhat magic, but at least consistent in the params):

  ```ruby
  Etsource::Inputs.new(etsource, :input).import
  Etsource::Inputs.new(etsource, :initializer_input).import
  ```

app/models/initializer_input.rb
```diff
@@ -0,0 +1,19 @@
+class InitializerInput
+  include Input::Common
+
+  ATTRIBUTES = [*Atlas::InitializerInput.attribute_set.map(&:name), :key]
+
+  attr_accessor *ATTRIBUTES
+
+  def self.get(key)
```
antw: Input and Iniitalizer input both define `.get` and `.fetch` and have identical implementations. Can they exist in `Input::Common` instead?

If you're finding that the `get` and `set` from InMemoryRecord are overwriting the ones from `Common`, you can instead include IMR into Common like so:

```ruby
class Input
  module Common
    extend ActiveSupport::Concern
    include InMemoryRecord

    included do
      include CommandAttributes
    end

    # ...
  end
end
```

(including a module as part of the `included` block will include it *after* Common; including it as part of Common itself will do so *before* Common).

|    | `Input.ancestors` |
| --- | --- |
| Before | Input, ActiveModel::Validations::HelperMethods, ActiveSupport::Callbacks, ActiveModel::Validations, CommandAttributes, **InMemoryRecord, Input::Common**, ... |
| After | Input, ActiveModel::Validations::HelperMethods, ActiveSupport::Callbacks, ActiveModel::Validations, CommandAttributes, **Input::Common, InMemoryRecord**, ... |

In fact, you might prefer to do that for the other two modules as well?

```ruby
class Input
  module Common
    extend ActiveSupport::Concern
    extend ActiveModel::Naming

    include InMemoryRecord
    include CommandAttributes

    included do
      attr_accessor :file_path
    end

    # ...
  end
end
```

app/models/etsource/inputs.rb
grdw: We can also do this:

```ruby
Atlas::InitializerInput.to_s.gsub(/.*\:\:/, '').constantize # => InitializerInput
```

But that seems like a really horrible unstable solution 😆

antw:
> Move 'get' and 'set' methods to common base module
> Clearify initialization of inputs and initializer inputs

@grdw Looks good to me! Happy for this to be merged?

grdw:
> Happy for this to be merged?

:shipit: Yes please!

Issue #918, opened 2017-04-20
Scaled derived regions have incorrect change in CO2, bio-footprint


antw:
For example:

#### Drenthe, full-size:

![drenth-full](https://cloud.githubusercontent.com/assets/4383/25231012/1217b2b0-25ce-11e7-99a2-9f20ebd26121.png)

#### Drenthe, 1000 residences (0.47% original size):

![drenthe-scaled](https://cloud.githubusercontent.com/assets/4383/25231025/1de14764-25ce-11e7-8b05-e656fffd5597.png)

#### Drenthe, 7,000,000 residences (~3000% original size):

![drenthe-scaled2](https://cloud.githubusercontent.com/assets/4383/25231198/dff3dbbe-25ce-11e7-99e8-ff3864354952.png)

"Normal" regions (non-derived) do not appear to be affected.

---

Including @ChaelKruip.

ChaelKruip:
Could this be because the reference value (1990) is not scaled properly?

antw:
> Could this be because the reference value (1990) is not scaled properly?

I think it goes beyond that. Looking at the bio-footprint query, there's a calculation of the [area footprint](https://github.com/quintel/etsource/blob/129c4526011d9ede4a99e4559d96cdc3c9cc487e/gqueries/general/area/bio_footprint.gql#L1) which looks at the [demand of converters in the "bio_footprint_calculation" group](https://github.com/quintel/etengine/blob/8a32f96a02a494cf10259c499aabdcc144fdbfdd/app/models/qernel/graph_api.rb#L58). I ran this query:

```ruby
SUM(V(GROUP(bio_footprint_calculation), demand))
```

In a full-size Drenthe scenario, their demand is `	
3,981,639,219` (3.9B). In a scenario scaled to 1000 households with all sectors enabled, their demand is `4,873,200,966` (4.8B)!

Here are the CO2 emissions compared. It looks like the 1990 figure is scaled correctly; but the actual demands are far too high:

| Full-size | Scaled to 1000 households |
| --- | --- |
| ![co2-dr-full](https://cloud.githubusercontent.com/assets/4383/25231994/c985c498-25d1-11e7-91ea-0e1251ba2fa7.png) |  ![co2-dr-scaled](https://cloud.githubusercontent.com/assets/4383/25232000/d0d27c00-25d1-11e7-8f59-794e0738e6c9.png) |

Strangely, industry CO2 does seem to be scaled, while everything else isn't.

[...]

PR #919, opened 2017-04-24
Scale after the initializer inputs have run

grdw:
Fixes: #918


grdw:
I'm not sure how to properly test this with specs or where this should be tested (which `spec` files).

PR #945, opened 2017-08-31
Prototype for the new initializer inputs

grdw:
Related to https://github.com/quintel/atlas/pull/100


app/models/qernel/area.rb
```diff
@@ -32,6 +32,14 @@ def inspect
       "<Area #{area_code}>"
     end
 
+    def initializer_inputs
+      begin
+        Atlas::Dataset::Derived.find(area_code).initializer_inputs
```
grdw: @antw I know that this is not the way to do this, but I didn't quiet understand how else to approach this problem. Could you maybe show me what a nicer way would be? 😄

antw: On `master`, the data should already be available on the `Qernel::Area` with the same name as in Atlas (`init`).

```ruby
gql = Scenario.default(area_code: 'noorderplantsoen').gql

gql.future.graph.area.init
# => { :households_useful_demand_cooking_per_person=>717610.2,
#      :households_cooker_halogen_electricity_share=>33.6,
#      ...
```

That's because Atlas attributes are automatically copied across to the `Qernel::Area`:

https://github.com/quintel/etengine/blob/9d21f2fcee31818785092129eff782c1dca4ff16/app/models/qernel/area.rb#L7-L15

I think you can achieve the same thing for your new method like so:

```ruby
ATTRIBUTES_USED = 
  [Atlas::Dataset::Full, Atlas::Dataset::Derived]. 
    .map { |klass| klass.attribute_set.map(&:name) }. 
    flatten +
    [:initializer_inputs] -
    [:id, :parent_id] 
```

grdw:
@antw one thing. Could you hold this off for merging for a few moments. ETsource probably needs a lot of work before this can properly work. Merging this now will probably disable all the local datasets.

app/models/etsource/dataset/import.rb
```diff
@@ -86,7 +86,18 @@ def load_graph_dataset
     #
     # Returns a hash.
     def load_region_data
-      { area_data: @atlas_ds.to_hash }
+      { area_data: @atlas_ds.to_hash.merge(
```
antw: This feels like something that Atlas should be doing itself.

app/models/graph_updater.rb
```diff
@@ -0,0 +1,29 @@
...
+    when 'unit_setter'   then set_unit(graph, *args)
+    else
+      raise ArgumentError, "no such method to update the graph #{ method }"
+    end
+  end
+
+  def self.demand_setter(graph, node_name, value)
+    node = graph.converter(node_name)
+
+    node.demand = value
```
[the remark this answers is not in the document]
grdw: I'll try and see what happens! 👍

grdw:
Thanks for the comments @antw

grdw:
> Merging this now will probably disable all the local datasets.

@antw this is no longer an issue. Merge at will 👍

app/models/graph_updater.rb
```diff
@@ -0,0 +1,53 @@
...
+    link = graph.link(edge_name)
+
+    link.share = value
+  end
+
+  def self.conversion_setter(graph, slot_name, value)
+    node_name, carrier = slot_name.to_s.split('@')
+
+    node = graph.converter(node_name)
+    slot = node.slots.detect{ |s| s.carrier.key == carrier.to_sym }
```
antw: A small thing I noticed here: the format for conversion setters is `node_key@carrier`. However this assumes that the converter has only one matching slot when in fact it could have two: input and output.

If a node has an `electricity` input, and `electricity` and `loss` outputs, what does setting `node@electricity: 0.8` do? It could set either the input or output to 0.8. Atlas specifies slot keys as having a `+` or `-` after the node key, defining whether it is an input or output slot: `node_key[+-]@carrier`.

* `+`: Input slot, adds (+) energy to the node
* `-`: Output slot, takes away (-) energy from the node

For example:

```ruby
node = Atlas::Node.find(:households_flexibility_p2p_electricity)
(node.in_slots.to_a + node.out_slots.to_a).map(&:key)
```

* `households_flexibility_p2p_electricity+@electricity`
* `households_flexibility_p2p_electricity-@loss`
* `households_flexibility_p2p_electricity-@electricity`

Perhaps initializer methods could use the same format?

PR #963, opened 2017-11-13
Deprecate initializer inputs

grdw:
Part of https://github.com/quintel/atlas/pull/106

### DIFF
--- a/app/models/gql/gql.rb
+++ b/app/models/gql/gql.rb
@@ -285,14 +285,21 @@ module Gql
 
     def update_present
       instrument('gql.performance.present.update_present') do
-        scenario.inputs_present.each { |input, value| update_graph(present, input, value) }
+        scenario.inputs_present.each do |input, value|
+          update_graph(present, input, value)
+        end
       end
     end
 
     def update_future
       instrument('gql.performance.future.update_future') do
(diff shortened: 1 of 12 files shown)


app/models/qernel/graph.rb
```diff
@@ -412,10 +411,6 @@ def graph_query
   end
 
   def initializer_inputs
-    decorated_inputs.sort_by { |input, _| [-input.priority, input.key] }
```
antw: Does this mean that init. inputs are no longer sorted by priority, or is it already done elsewhere?

app/models/gql/gql.rb
```diff
@@ -285,14 +285,21 @@ def calculate_graphs
 
     def update_present
       instrument('gql.performance.present.update_present') do
-        scenario.inputs_present.each { |input, value| update_graph(present, input, value) }
```
antw: I'm nitpicking (again) and it doesn't matter now, but can I suggest that in the future we try to keep code-style/whitespace changes out of pull requests for new features? (same for [topology.rb](https://github.com/quintel/etengine/pull/963/files#diff-547f837cee3bb15737d186fdd290d314))

There style changes are A Good Thing™, but they're not related to deprecating initializer inputs.

PR #968, opened 2017-11-16
Public transport

grdw:


db/migrate/20171116110243_migrate_existing_scenarios_for_public_transport_changes.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    # Pruned input keys
+    removed = [
+      'transport_useful_demand_ship_kms',
+      'transport_useful_demand_trains',
+      'transport_useful_demand_planes',
+      'transport_truck_using_compressed_natural_gas_efficiency'
+    ]
+
+    scenarios = Scenario.where("`user_values` IS NOT NULL AND `balanced_values` IS NOT NULL AND `user_values` != '--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess {}\n' AND `balanced_values` != '--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess {}\n'")
```
antw:
```sql
`user_values` != '--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess {}\n' AND
`balanced_values` != '--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess {}\n'")
```

Will this not miss scenarios where `user_values` has inputs and `balanced_values` is empty, and vice-versa?

---

```sql
SELECT count(id) FROM scenarios
-- 598742

SELECT count(id) FROM scenarios
WHERE
	`user_values` IS NOT NULL AND
	`balanced_values` IS NOT NULL AND
	`user_values` != '--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess {}\n' AND
	`balanced_values` != '--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess {}\n'
-- 77119

SELECT count(id) FROM scenarios
WHERE
	( `user_values` IS NOT NULL OR
		`balanced_values` IS NOT NULL )
	AND
	( `user_values` != '--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess {}\n' OR
		`balanced_values` != '--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess {}\n' )
-- 414564
```

db/migrate/20171116110243_migrate_existing_scenarios_for_public_transport_changes.rb
```diff
@@ -0,0 +1,98 @@
...
+
+        # Actual migration
+        scenario.public_send("#{ inputs_attribute }=", migrated)
+      end
+
+      if scenario.valid?
+        scenario.save
+        bar.increment!
+      else
+        a = "#{ DateTime.now } - #{ scenario.id } can't be migrated because '#{ scenario.errors.messages }'\n"
```
antw: Are there any scenarios that cannot be migrated on the staging or production DBs?

I tend to think this is a fatal error and the migration should exit, but if the failing validation is unrelated to the new/renamed inputs you can just run `scenario.save(validate: false)` safe in the knowledge that the migration is working as expected. If the record was *already* invalid, I expect it is better for it to be invalid with the correct inputs than to be invalid with the *old* inputs.

grdw: All the errors are related to non-existing `area_code`.s

grdw:
> This is still really slow for me (38 minutes against the staging DB), but partially that's due to deserializing YAML for 400k rows.

Hmm.. I added a filter to check if any of the keys - that need migrating - are even present to begin with. I don't know about the time, but it still felt very slow.

antw:
> check if any of the keys - that need migrating - are even present to begin with. I don't know about the time, but it still felt very slow.

Does this improve performance for you? I tried this myself last week and found that it performed worse than the original on my laptop.

grdw:
@antw - as discussed - we use your migration script. Could you approve this pull request. Using my admin "Merge" button feels a little gross.

antw:
> Could you approve this pull request.

I think I pushed some stuff with that migration that I didn't mean to: `(relevant_keys & migrated.keys)`.

One moment and I'll clean it up.

antw:
Fixed my mistakes. Annoyingly GitHub won't let me assign @grdw to review. @grdw: I'll approve this and if you're happy with it, go ahead and merge. :+1:

PR #978, opened 2017-12-14
Clear region list prior to calculating datasets

### LINKED ISSUE #977: New datasets in ETSource aren't calculated and messagepacked automatically
Everytime a new dataset is imported from ETSource - the `deploy:calculate_datasets` task doesn't seem to run (or people are impatient 😆). 

I always seem to have to do this manually somehow 🤔 which fixes the 'ETmodel' from freezing up.

antw:
Importing a new ETSource commit would attempt to reload the Atlas data before the Rails/in-memory caches had been expired (largely to prevent old Atlas data from re-populating the caches).

This commit adds `NastyCache.delete` and will remove the cached list of regions prior to reloading Atlas data, thus ensuring that newly-added or -enabled datasets are calculated.

Closes #977

### DIFF
--- a/app/models/etsource/dataset.rb
+++ b/app/models/etsource/dataset.rb
@@ -24,7 +24,9 @@ module Etsource
       ::Etsource::Dataset::Import.new(country).import
     end
 
-    def self.region_codes
+    def self.region_codes(refresh: false)
+      NastyCache.instance.delete('region_codes') if refresh
+
       NastyCache.instance.fetch('region_codes') do
         Atlas::Dataset.all
           .select { |dataset| dataset.enabled[:etengine] }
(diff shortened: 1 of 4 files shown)


app/models/etsource/dataset.rb
```diff
@@ -24,7 +24,9 @@ def import
       ::Etsource::Dataset::Import.new(country).import
     end
 
-    def self.region_codes
+    def self.region_codes(refresh: false)
+      NastyCache.instance.delete('region_codes') if refresh
```
grdw: I'm not a really big fan of putting if-statements after a line (since it makes debugging a little harder), but considering that it's a relatively short piece of code I'm okay with it. The alterantive is equally ugly though 😆 :

```ruby
if refresh
  NastyCache.instance.delete('region_codes')
end
```

antw: I've no strong opinion on this myself; just [following the style guide](https://github.com/bbatsov/ruby-style-guide#if-as-a-modifier).

PR #979, opened 2017-12-19
Created new update type

nairnamratha:
New update type: UPDATE_WITH_NUMBER functions as the default update when the update type is set to other other types.

Created with reference to the changes proposed here : https://github.com/quintel/etsource/pull/1492

### DIFF
--- a/app/models/gql/runtime/functions/update.rb
+++ b/app/models/gql/runtime/functions/update.rb
@@ -128,6 +128,12 @@ module Gql::Runtime
         end
       end
 
+      def UPDATE_WITH_NUMBER(*value_terms)
+        update_something_by(*value_terms) do |original_value, input_value|
+          input_value
+        end
+      end
+
       # @private
       def update_something_by(*value_terms)
         input_value_proc = value_terms.pop


grdw:
Just want to include @antw and get his opinion - if time permits. If not, please do tell.

I don't like the `UPDATE_WITH_NUMBER` name. It's exactly the same as an `UPDATE` with the `:absolute` strategy. So my suggestion would be to name it:

`UPDATE_ABSOLUTE`.

Also it would be nice to include some comment in the code explaining what it does, much like the other methods. I added those changes [here](https://github.com/quintel/etengine/compare/master...nami-update-changes). If that looks okay to you?

antw:
> I don't like the UPDATE_WITH_NUMBER name. It's exactly the same as an UPDATE with the :absolute strategy.

I don't think the new method is necessary – please correct me if I'm wrong. [The default strategy *is* `:absolute`](https://github.com/quintel/etengine/blob/9df4ca30ff6098fa9925271e93fb1793d261f950/app/models/gql/runtime/functions/update.rb#L187-L206), so this is identical to just calling `UPDATE`.

nairnamratha:
But how do I call UPDATE when the update type is set to something else, for example %y ?

grdw:
@antw to explain the situation a little. It's related to: https://github.com/quintel/etsource/pull/1492

In that PR there are changes to a slider which performs a series of `UPDATE` statements. The last one is a correction for the things changed with those `UPDATE` statements. The sliders `update_type` is `%y` and the correction needs to be updated absolutely.

antw:
> But how do I call UPDATE when the update type is set to something else, for example %y ?
> The sliders update_type is %y and the correction needs to be updated absolutely.

Good questions. In that case I guess the `UPDATE` is being forced to `relative_per_year` when you explicitly don't want that? I agree that `UPDATE_ABSOLUTE` is a better name.

I'm not so keen on adding a new method just for one query, but I'm struggling to come up with a better idea.

grdw:
> I'm not so keen on adding a new method just for one query, but I'm struggling to come up with a better idea.

Same problem over here 😓

antw:
> Edit: I did had another idea, where the UPDATE method could take a second argument called force_strategy to set the strategy manually. But it doesn't feel 'better'.

Yeah, I agree. That might be quite fragile since `UPDATE` tries (too) hard to be flexible with arguments. It dates back to the Ruby 1.8 days, but in newer Rubies (> 2.0?) it's method signature would be:

```ruby
def update (*things_to_update, attribute_name, value)
  # ...
end
```

... where `things_to_update` is any number of objects (or arrays of objects) to be updated. Therefore, it's kind of hard to add new arguments and options without risking breaking some ETSource inputs

A longer-term alternative might be to do away with `- update_type` altogether, add some new `UPDATE_*` methods, and be more explicit in the queries:

* `UPDATE` (today: update_strategy=absolute)
* `UPDATE_WITH_FACTOR` (as today)
* `UPDATE_RELATIVE` (today: update_strategy=relative)
* `UPDATE_RELATIVE_PER_YEAR` (today: update_strategy=relative_per_year)
* ...

Maybe that's a nice-to-have for next year.

For the moment, perhaps we should just push ahead with `UPDATE_ABSOLUTE`?

grdw:
> Maybe that's a nice-to-have for next year.

Yes. Are you noting this one down or should I aswell?

> For the moment, perhaps we should just push ahead with UPDATE_ABSOLUTE?

Yes, I guess that would help @nairnamratha for now.

antw:
> Yes. Are you noting this one down or should I aswell?

[I added a reminder in Basecamp](https://3.basecamp.com/3797409/buckets/4288547/todos/826495618) to remind us. I think it's very low-priority; just something to explore when we have a day or two free.

PR #980, opened 2017-12-19
Nami update changes

grdw:
As follow up from https://github.com/quintel/etengine/pull/979


grdw:
@nairnamratha could you validate locally  that this branch is functional with your ETSource PR https://github.com/quintel/etsource/pull/1492?

PR #981, opened 2017-12-20
Adjust buildings solar PV production

antw:
Changes the amount of solar PV installed on buildings so as to ensure that solar production of affected scenarios remains constant after the changes in quintel/etsource#1492.

I expect this should be merged simultaneously with quintel/etsource#1492.

—

*Including @nairnamratha*

### DIFF
--- a/db/migrate/20171220132741_update_buildings_solar_pv_production.rb
+++ b/db/migrate/20171220132741_update_buildings_solar_pv_production.rb
@@ -1,0 +1,58 @@
+class UpdateBuildingsSolarPvProduction < ActiveRecord::Migration
+  INPUTS = %w[
+    buildings_solar_pv_solar_radiation_market_penetration
+    number_of_buildings
+  ].freeze
+
+  SOLAR_KEY = :buildings_solar_pv_solar_radiation_market_penetration
+
+  def up
+    update_each_scenario do |scenario|
+      scenario.user_values[SOLAR_KEY] /= growth_factor(scenario)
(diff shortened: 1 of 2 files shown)


db/migrate/20171220132741_update_buildings_solar_pv_production.rb
```diff
@@ -0,0 +1,58 @@
...
+    update_each_scenario do |scenario|
+      scenario.user_values[SOLAR_KEY] *= growth_factor(scenario)
+    end
+  end
+
+  private
+
+  def update_each_scenario
+    # Only update protected scenarios or those created in the past month (older
+    # unsaved scenarios are likely abandoned). Skip mturk scenarios.
```
grdw: Just out of curiosity, but is this going to be the default behavior when migrating scenarios? To support only 'protected' scenarios and one's which are created in the last month? If so, that would be amazing! 👍

grdw:
Just ran the migration on my local machine with a fairly recent staging database and everything seems to work 👍

db/migrate/20171220132741_update_buildings_solar_pv_production.rb
antw: Yeah, I think it makes sense. Old unsaved scenarios are unreachable in the ETM (unless you've written down the ID somewhere) so I don't think there's much advantage to migrating them.

Unless there are objections, I think we should do this for all scenario migrations in the future.

PR #985, opened 2018-01-05
Add migration for CO2 emissions

grdw:
Migrated scenarios count: 3672

Time between each 50 scenarios:

```
2018-01-05 17:01:06 +0100 | 0 -> 114947
2018-01-05 17:01:53 +0100 | 50 -> 116948
```

It's about a minute per 50. So 3000 / 60 =~ 50 minutes before this migration is finished.

### DIFF
--- a/db/migrate/20180104152128_correct_co2_emission_sliders.rb
+++ b/db/migrate/20180104152128_correct_co2_emission_sliders.rb
@@ -1,0 +1,34 @@
+class CorrectCo2EmissionSliders < ActiveRecord::Migration
+  KG_PER_MJ_IN_G_PER_KWH = 3600.0
+
+  def change
+    scenarios = Scenario.where(
+      '(protected = ? OR created_at >= ?) AND source != ? AND title != ?',
+      true, 1.month.ago, 'Mechanical Turk', 'test'
+    ).order(:area_code)
+
+    count = 0
+    puts "Need to migrate: #{ scenarios.count } scenarios"
+
+    scenarios.find_each do |scenario|
+      next unless scenario.valid?
+
+      if count % 50 == 0
+        puts "#{ Time.now } | #{ count } -> #{ scenario.id }"
+      end
+
+      gql = scenario.gql
+
+      scenario.user_values.merge!(
+        co2_emissions_of_imported_electricity_present: KG_PER_MJ_IN_G_PER_KWH *
+          gql.present.query(:average_co2_emissions_of_produced_electricity),
+        co2_emissions_of_imported_electricity_future: KG_PER_MJ_IN_G_PER_KWH *
+          gql.future.query(:average_co2_emissions_of_produced_electricity)
+      )
+
+      scenario.save
+
+      count += 1
+    end
+  end
+end
--- a/db/migrate/20180104152128_correct_co2_emission_sliders.rb
+++ b/db/migrate/20180104152128_correct_co2_emission_sliders.rb
@@ -1,7 +1,7 @@
-class CorrectCo2EmissionSliders < ActiveRecord::Migration
+class CorrectCo2EmissionSliders < ActiveRecord::Migration[5.1]
   KG_PER_MJ_IN_G_PER_KWH = 3600.0
 
-  def change
+  def up
     scenarios = Scenario.where(
       '(protected = ? OR created_at >= ?) AND source != ? AND title != ?',
       true, 1.month.ago, 'Mechanical Turk', 'test'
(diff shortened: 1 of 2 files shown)


db/migrate/20180104152128_correct_co2_emission_sliders.rb
```diff
@@ -0,0 +1,34 @@
...
+    puts "Need to migrate: #{ scenarios.count } scenarios"
+
+    scenarios.find_each do |scenario|
+      next unless scenario.valid?
+
+      if count % 50 == 0
+        puts "#{ Time.now } | #{ count } -> #{ scenario.id }"
+      end
+
+      gql = scenario.gql
```
grdw: @antw is there anything you can do that can make this migration any quicker? 😓

antw: > @antw is there anything that we can do that can make this migration any quicker? 😓

What is the purpose of this migration?

I understand the *what* (set CO2 emissions of imported electricity), but I don't get the *why*. As far as I know, in the model currently this will be zero; why do we want to update existing scenarios to have a non-zero value?

---

As for making it quicker, I think my solution would be to replace the exact calculation of CO2 (which requires that you calculate the scenario using `scenario.gql`) with an approximation. For example, instead take the CO2 emissions from a default scenario and use that instead of calculating it for every saved scenario. If it's a requirement for every scenario to have the exactly-correct value, that might not be an acceptable solution.

grdw: @MartLubben actually wants to hold this off, since there's a bit of discussion about: 

> What is the purpose of this migration?

^ that exactly.

So I'm going to close this for now. If it's needed, I'll push the branch again and add an explanation about the what and why.

grdw:
@MartLubben What actually happened to this migration problem? Is this still needed in the nearby future?

grdw:
> Before the sliders which set the CO2 of imported electricity, the CO2 was set by the average produced CO2 of the production park (in the future). Because all production parks massively differ, the average CO2 of electricity and thereby the total CO2 of imported electricity massively differs.

I'm really sorry but I have no idea what this says. 

As far as I understand the issue, it is the following: a serie of sliders influences the CO<sub>2</sub> of imported electricty. This is a value which needs to be initialized per region (I can remember something with `co2_conversion_per_mj` of `imported_electricity` and something about using the FCE functionality?). The graph needs to be adjusted to that value (so specific nodes in specific places need a demand, a share, etc.) in order for the graph to *load* correctly for every scenario.   

The only rule I'm (and @antw aswell - please correct me if this is not the case) trying to enforce here is: that the graph and the dataset should initialize correctly. It shouldn't be initialized with the idea that a slider setting should be set, to correct your 'scenario'. In other words: ETEngine should be spoon-fed the correct values to setup the correct CO<sub>2</sub> values for imported electricity.

It was to my understanding that this was being figured out, how to setup the graph properly. But I guess that's not going as planned? Or what's the status of that?

MartLubben:
@grdw . I am sorry that my communication about this topic is somewhat unclear, also because my thoughts were changing a couple of times. I will try to share my current thoughts in a more clear way in this message. If it is still unclear I would like to do it in person soon.

First:
I thought that we should keep the dashboard outcomes (CO2, Costs, Renewability) of some protected scenario's somewhat stable after any change in the ETM. See discussion: https://3.basecamp.com/3797409/buckets/4288548/messages/843023681 
 And specific decision: 
> Bug fix migrations use the above criteria (presets, scenarios modified in the past month, protected scenarios, etc).
> Migrations to maintain outcomes apply to presets and protected scenarios only.
 
Second:
The dashboard output (CO2) of those protected scenario's are at the moment significantly changed. At this moment a fixed value is chosen for the present and future. Using the FCE-like functionality will only offer a handful of fixed values. However the values in the scenarios can be somewhat between 0 and 1000 g/kWh. Those values were (before) set invisibly by taking the average (CO2) of the own production park. Because every scenario has a different production park, these values differ a lot. 

In conclusion, I do think the migration needs to be done for all scenario's were the dashboard outcomes should be stable. also, I do think that this group can be (much) smaller than **all** protected scenario's. However that is another discussion. @grdw @antw should I reopen the migration discussion in the basecamp and share this thought?

grdw:
> I thought that we should keep the dashboard outcomes (CO2, Costs, Renewability) of some protected scenario's somewhat stable [..]

Yes .... and no. I think that the dashboard values should be stable for some scenario's to a certain extent. Technology around us changes and I think we shouldn't be conservative - just for the sake of some dashboard values. I get that there are reports being written and that values for *those scenarios* should remain somewhat consistent, but only those scenarios. 

> should I reopen the migration discussion in the basecamp and share this thought?

Yes! What is a 'protected' scenario - other than a 'saved' scenario? Currently - technically - there's no difference 😅

PR #988, opened 2018-01-10
Added flexibility to technical and financial parameters of tooltips

marliekeverweij:
I have just added an extra section to the file that generates the technical financial parameters in the tooltips. To be clear, this is what I want in the tooltips of p2l:

screen shot 2018-01-03 at 14 36 39

@joris, could you check if the right attributes are included?
@grdw, could you check if the syntax is correct?

Thanks!

### DIFF
--- a/app/models/api/v3/converter_presenter_data.rb
+++ b/app/models/api/v3/converter_presenter_data.rb
@@ -185,6 +185,52 @@ module Api
         }
       }
 
+      # If the converter belongs to the :cost_electricity_production group then
+      # add these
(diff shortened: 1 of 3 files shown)


grdw:
> @grdw, could you check if the syntax is correct?

This might seem like a silly answer, but you can check yourself if the syntax is correct which is through running: `(bundle exec) rspec` in the `etengine` folder. If it doesn't crash the second it starts, the syntax is correct 👍 

I am going to check if it's correct semantically and in line with the [code-styling](https://github.com/bbatsov/ruby-style-guide#collections). That's all 😅 

Conceptually @jorisberkhout is assigned I see. 👍

PR #989, opened 2018-01-17
Add migration for fuel prices

grdw:
@nairnamratha would you be so kind as to run this migration locally. 

So check out this branch, do `rake db:migrate` and validate if a scenario with the 'old price' settings is being migrated correctly. 

I've tested this and it seems to work on my machine, but it's nice to get a second validation.


grdw:
@nairnamratha do you know by when this is okay to be merged?

nairnamratha:
Looks like its ready to be merged @grdw ! Tested it locally and everything seems to be migrated correctly.

grdw:
Could you review it positively than 😅 - by pushing "Add review" and say "Approve".

PR #994, opened 2018-01-24
Migration for co2 emissions

grdw:
@MartLubben could you validate that this migration is working out as expected. You can do this as follows:

1. Checkout this branch locally
2. Make sure you have a backup of your local `etengine` database
3. Run the migration
4. Cherry-pick some important scenarios and validate if their outcome for the CO<sub>2</sub> dashboard item is still equal to the scenario on the beta server.

If you need help with this I'm happy to provide it.


grdw:
Sorry to close this again. But it's currently not possible to migrate scenarios which depend on a previously defined gquery. 😞 I found this out the hard way with @nairnamratha with the fuel prices migration. 

The reason for this is that a dataset is not accessible from a migration:

screen shot 2018-01-30 at 15 49 47

I can pre-calculate some values, but that's all there's to it. Will do that when time permits.

PR #999, opened 2018-02-02
Permit correct parameters for converters controller

grdw:
@antw I don't have a strong preference to use:

This...

```ruby
keys = params.require(:keys).permit!.to_h
```

... over my fix. So feel free to overwrite it if you feel like it 😆

### DIFF
--- a/app/controllers/api/v3/converters_controller.rb
+++ b/app/controllers/api/v3/converters_controller.rb
@@ -31,7 +31,7 @@ module Api
       #   "keyN": [ .. ]
       # }
       def stats
-        keys = params.require(:keys)
+        keys = permitted_params.to_h.fetch(:keys)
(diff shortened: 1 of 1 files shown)


app/controllers/api/v3/converters_controller.rb
```diff
@@ -31,7 +31,7 @@ def show
       #   "keyN": [ .. ]
       # }
       def stats
-        keys = params.require(:keys)
+        keys = permitted_params.to_h.fetch(:keys)
```
grdw: The `#to_h` is needed or else it will throw this really long [Deprecation message in the logs](https://stackoverflow.com/questions/38710904/how-do-i-resolve-the-deprecation-warning-method-to-hash-is-deprecated-and-will).

PR #1001, opened 2018-02-13
Add storage_volume to whitelisted methods for ETMoses

grdw:
Part of https://github.com/quintel/etmoses/pull/1616 *

### DIFF
--- a/app/models/api/v3/converter_stats_presenter.rb
+++ b/app/models/api/v3/converter_stats_presenter.rb
@@ -13,12 +13,16 @@ module Api
         full_load_hours: nil,
         fixed_operation_and_maintenance_costs_per_year: nil,
         variable_operation_and_maintenance_costs_per_full_load_hour: nil,
-        variable_operation_and_maintenance_costs_for_ccs_per_full_load_hour: nil,
+        variable_operation_and_maintenance_costs_for_ccs_per_full_load_hour: nil
       }
 
       ATTRIBUTES = {
         heat_output_capacity: nil,
         marginal_heat_costs: nil,
+
+        :'storage.volume' => -> conv {
+          conv.storage.volume
+        },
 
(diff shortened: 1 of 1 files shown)


grdw:
Semaphore seems to break because of unknown reasons. Could you maybe look what the problem might be @antw ?

app/models/api/v3/converter_stats_presenter.rb
```diff
@@ -13,13 +13,17 @@ class ConverterStatsPresenter
...
         variable_operation_and_maintenance_costs_per_full_load_hour: nil,
-        variable_operation_and_maintenance_costs_for_ccs_per_full_load_hour: nil,
+        variable_operation_and_maintenance_costs_for_ccs_per_full_load_hour: nil
       }
 
       ATTRIBUTES = {
         heat_output_capacity: nil,
         marginal_heat_costs: nil,
 
+        :'storage.volume' => -> conv {
```
antw: Personally I think this should be "storage_volume" since this is part of ETE's public API. `volume` being a property of `storage` is an implementation detail.

grdw: Yes, I guess this has to do with old naming in ETMoses. I'll see to it to change it, as it becomes a bit of problem now ETMoses. 

Related to: https://github.com/quintel/etmoses/issues/1613

antw: > Yes, I guess this has to do with old naming in ETMoses.

Good point. Perhaps ETMoses was requesting the `storage` attribute, which is a hash that `volume`? That would fit with the storage attribute already being whitelisted: https://github.com/quintel/etengine/blob/1607a71990fa3a154145b12e4ec02542a35d3a82/app/models/api/v3/converter_stats_presenter.rb#L9

PR #1010, opened 2018-03-23
Hydrogen converter positions

marliekeverweij:
Converter positions of new hydrogen nodes and 

! I do not know if first the hydrogen ETSource branch should be merged before merging this ETEngine branch!

### DIFF
--- a/app/models/qernel/converter_api.rb
+++ b/app/models/qernel/converter_api.rb
@@ -145,6 +145,28 @@ class ConverterApi
     end
   end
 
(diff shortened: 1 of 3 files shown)


antw:
> I do not know if first the hydrogen ETSource branch should be merged before merging this ETEngine branch

I don't think it matters in which order the branches are merged, but this one will need to be merged before the hydrogen commit is *imported* in ETEngine.

I'll approve the PR but leave it up to you to press "Merge" when everything else is ready.

PR #1015, opened 2018-04-17
Migrate hydrogen slider from shares to number of units

grdw:


grdw:
@antw 

> Does the "pre_migration_script.rb" have to be run first? The migration will have to call that script in order to build the JSON during a deploy... could we just move the contents of the pre-migration script into the migration itself?

No. You can't access the `gql` bit of a scenario during a migration. I'm using the `rails runner` right before the migration and use the `data.<env>.json` to update the scenarios.

grdw:
> I remember this being said before, but I'm not sure why it is the case. Was this because the graph structure had changed in a way that prevented the migration from working correctly? How does running the pre-migration script separately avoid this?

Well the pre-migration script doesn't run automatically. It runs manually by copying it over to the server and using the server's `rails runner`. This will generate the data needed for the actual migration. 

This way you won't have to access the `gql` instance. I think the problem lies within the fact that you can't access the `.pack` files from the migration as they aren't there yet at the time the migration is running. I think the deploy goes like:

``` 
1. copying over all files
...
...
9. running all migrations
...
...
15. Calculating the datasets
```

antw:
> I think the problem lies within the fact that you can't access the datasets .pack files from the migration as they aren't there yet at the time the migration is running. I think the deploy goes like: [...]

Thanks for explaining, that makes a lot of sense. Indeed the deploy tasks are in that order:

* [...]
* 00:00 deploy:assets:precompile
* 00:00 deploy:assets:backup_manifest
* **00:00 deploy:migrate**
* **00:00 deploy:migrating** ← migrations
* **00:00 deploy:etsource** ← check out and calculate ETSource
* 00:00 deploy:symlink:release
* 00:00 memcached:flush
* 00:00 deploy:log_revision
* 00:00 puma:restart

We could run the ETSource task before the migrations, but I'm not sure if Rails will just throw an error if there are pending migrations?

---

For this PR, maybe the part of the migration which uses the JSON should be *in* the pre-migration script? Otherwise the migration will run automatically against the empty JSON file. We would just have to remember to run the script manually after the deploy. Does that seem reasonable?

antw:
> For this PR, maybe the part of the migration which uses the JSON should be in the pre-migration script? Otherwise the migration will run automatically against the empty JSON file. We would just have to remember to run the script manually after the deploy. Does that seem reasonable?

It occurs to me that this won't work as the old inputs – which set the shares of these technologies – won't exist anymore. It's also likely that it's too late for this to work on staging since those inputs are presumably long gone already. 😬

H2 is getting deployed this morning, so I suggest that:

1. We build the JSON file on our own machines using a copy of the database(s) and the code / ETSource as it was before the H2 deploy. I'll grab copies of these and Git refs before the deploys happen.

2. Write a "post-deploy" script – which is not part of the migration – which we can upload and run (with the JSON) after the deploy.

grdw:
> Otherwise the migration will run automatically against the empty JSON file

For the staging environment this isn't terrible. The production environment however, should migrate with the data included. 

> It occurs to me that this won't work as the old inputs – which set the shares of these technologies [..] 

Yes, those are gone. So that's why we create the JSON file before the actual deploy of H2.

grdw:
> We build the JSON file on our own machines using a copy of the database(s) and the code / ETSource as it was before the H2 deploy. I'll grab copies of these and Git refs before the deploys happen.

So who does what now exactly?

PR #1020, opened 2018-06-26
Use imported_electricity cost to set MO import price

antw:
Refactors the ProducerAdapter into a separate DispatchableAdapter class,
and made ImportAdapter inherit from the Producer. Allows the import
interconnector to use the price of the imported electricity carrier in
order to behave like any other dispatchable.

Ref quintel/etmodel#2826

### DIFF
--- a/app/models/qernel/plugins/merit/import_adapter.rb
+++ b/app/models/qernel/plugins/merit/import_adapter.rb
@@ -1,64 +1,36 @@
 module Qernel::Plugins
   module Merit
-    class ImportAdapter < Adapter
+    # Implements behaviour specific to the import interconnector.
+    class ImportAdapter < ProducerAdapter
       def inject!
+        super
+
         elec_link = @converter.converter.output(:electricity).links.first
-        demand    = participant.production(:mj)
 
         if elec_link.link_type == :flexible
           # We need to override the calculation of the flexible link and set
           # set the demand explicitly.
-          elec_link.dataset_set(:value, demand)
+          elec_link.dataset_set(:value, @converter.demand)
           elec_link.dataset_set(:calculated, true)
         end
-
-        @converter.demand = demand
       end
 
       private
-
-      def producer_attributes
-        attrs = super
-
-        # The marginal cost of the import producer is 1 EUR more than the the
-        # most expensive participant.
-        #
-        # See https://github.com/quintel/merit/issues/143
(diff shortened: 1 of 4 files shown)


app/models/qernel/plugins/merit/import_adapter.rb
```diff
@@ -1,64 +1,36 @@
...
 
-        mo_converters.map do |conv|
-          begin
-            cost = conv.query.marginal_costs
-            cost.nan? || cost == Float::INFINITY ? 0.0 : cost
-          rescue
-            0.0
-          end
-        end.max
+      def flh_capacity
```
ChaelKruip: Here flh_capacity is defined as the electricity_output_capacity of the converter. This means that the demand of the node is based on its output. This is only correct if losses are zero. I don't understand this. Perhaps the name flh_capacity can be chosen to be more specific?

app/models/qernel/plugins/merit/producer_adapter.rb
```diff
@@ -42,23 +35,29 @@ def inject!
...
+        @converter.marginal_costs
+      end
+
+      def output_capacity_per_unit
+        @converter.electricity_output_conversion * @converter.input_capacity
+      end
+
+      # Internal: Capacity used to multiply full load seconds to determien the
+      # resulting annual demand of the producer.
+      def flh_capacity
```
ChaelKruip: For producers, the input capacity is used for the flh_capacity. Why is this different from how the import interconnector is treated?

app/models/qernel/plugins/merit/import_adapter.rb
antw: This is only used by the import interconnector. Sliders adjust the output capacity of this converter; it doesn't have an input capacity.

app/models/qernel/plugins/merit/producer_adapter.rb
antw: This is used to convert the FLH of the Merit producer back to a demand for ETEngine. It's not a very good explanation, but input_capacity is used because [that's what ETE has always used](https://github.com/quintel/etengine/commit/969da27ca80785c930199d4f85e03c49ba1fc115#diff-bd4da10f28ee1072afe37359a7f64ff7R62) (this is not a new addition; it's just moving what was already there to a new method so that the class is more flexible).

PR #1032, opened 2018-11-29
Add primary_demand_abroad method

antw:
* Calculates primary demand originating in nodes which are marked as "abroad", i.e. not part of the modelled region. A node can be marked as "abroad" in ETSource by adding the attribute...

  ```
  - abroad = true
  ```

  For example:

  ```ruby
  V(energy_import_greengas, primary_demand_abroad)
  # ... (returns zero for now as there are no "abroad" green-gas nodes)
  ```

  The standard `primary_demand` calculation will ignore energy originating in nodes flagged as abroad (essentially terminating the calculation when reaching a node whose inputs are all abroad).

* Adjusts recursive factor calculations to accept a "domestic" flag (enabled by default) setting whether to calculate a factor for domestic or "abroad" nodes. Recursive factor will filter out any values from the undesired localities.


antw:
@ChaelKruip I realised a minute too late that you already have a biomass branch: I'll add these changes to that branch.

PR #1034, opened 2018-12-18
Migrate some scenarios to use new dataset names

antw:
Some old datasets have been superseded by datasets created in ETLocal. This migrates scenarios using those datasets to use the new names, allowing for the removal of the old datasets from ETSource.

Ref quintel/etsource#1876


MartLubben:
@antw Great!

PR #1049, opened 2019-03-22
Migration for changes in solar park and coal-fired industry chp sliders

redekok:
Dataset migrations for two changes in slider inputs:

- The slider unit for `energy_power_solar_pv_solar_radiation` has changed from number of units to installed capacity. The slider settings are updated for this change. (https://github.com/quintel/etsource/pull/1961)
- The capacity of `industry_chp_ultra_supercritical_coal` has changed from 17.4 MW to 600 MW. The slider settings for this node (`number_of_industry_chp_ultra_supercritical_coal`) are scaled based on this increase in capacity. (https://github.com/quintel/etsource/pull/1964)

@antw Could you check the migration scripts?

Furthermore, when running the migration, the terminal output implies that zero scenarios are migrated when running the industry chp migration script. However, the scenarios (e.g., 1037603) do seem to be updated. @antw Do you know what is going on here?

Screenshot 2019-03-22 at 10 29 21
Screenshot 2019-03-22 at 10 35 10
Screenshot 2019-03-22 at 10 29 44
 
**Edit**: there might be no scenarios in the database in which the user has touched the slider (`number_of_industry_chp_ultra_supercritical_coal`) to change the number of units. This could be the reason for the fact that no scenarios have been migrated by this script.

@antw: Would you suggest to either remove the migration script for the changes regarding the `industry_chp_ultra_supercritical_coal` (since there do not seem to be any scenarios that should be migrated) or to keep it in this PR? In case of the latter, I am not sure how to check whether the script is correct. I've tried creating and saving a new scenario (on BETA) in which I've changed the slider setting but this scenario is of course not part of the database yet so it is not migrated by the script. What are your thoughts on this?


db/migrate/20190320155025_change_industry_chp_supercritical_coal_capacity.rb
```diff
@@ -0,0 +1,62 @@
+class ChangeIndustryChpSupercriticalCoalCapacity < ActiveRecord::Migration[5.1]
+  INPUTS = {
+    industry_chp_supercritical_coal: (17.3611111111111 / 600.0)
```
antw: The reason nothing happens in this migration is because the input/slider name is "number_of_industry_chp_ultra_supercritical_coal" not "industry_chp_supercritical_coal" (note the "number_of" prefix, and the addition of "ultra"). There is no input or converter called "industry_chp_supercritical_coal".

The first migration can omit the `number_of_` prefix because it also renames `number_of_` to `capacity_of_` (lines 15/16 and 34/35). That's not the case for this second migration: the full input name needs to be provided.

Once this is done, the migration should run as expected.

db/migrate/20190320145911_change_solar_park_inputs_to_capacity.rb
```diff
@@ -0,0 +1,67 @@
...
+    if scenario.user_values.key?(from)
+      scenario.user_values[to] = scenario.user_values.delete(from) * multiplier
+      true
+    end
+  end
+
+  # All protected scenarios, and any unprotected scenarios since Jan 1st 2018
+  # will be updated.
+  def scenarios
+    Scenario.migratable_since(Date.new(2018, 1, 1))
```
antw: Out of interest, why are all scenarios from the beginning of 2018 being migrated? Was this an intentional choice, or just copy/paste from another migration?

Normally we just do protected scenarios, and unprotected scenarios from the past month. In that case, you can just use `Scenario.migratable`:

```diff
  def scenarios
-   Scenario.migratable_since(Date.new(2018, 1, 1))
+   Scenario.migratable
  end
```

redekok: It's not really an intentional choice, actually. I've indeed copy/pasted the method from a similar migration for the hydrogen sliders that was performed recently. Anyways, thanks for the explanation. I'll adopt the suggested changes!

antw:
> Both migrations can be simplified quite a bit on account of them only updating a single input. I can show how to do this once the other comments have been addressed.

Here's how I would simplify the two migrations: https://gist.github.com/antw/8affd54a46d201d45dc36a94d7045ec7

The versions you've posted work fine, but they're a bit more complicated than they need to be, on account of the original H2 migration having to adjust several inputs. Since these migrations only adjust a single input, we can simplify them quite a bit:

1. We can get rid of the `collection.reduce` loop in each migration since we're only updating one input.
2. We can hard-code the input name and multiplier in a constant (the UPPERCASE variables at the top of the migration).
3. We can remove a couple of methods like `update_input` and `scenarios`.

db/migrate/20190320145911_change_solar_park_inputs_to_capacity.rb
michieldenhaan: > Normally we just do protected scenarios, and unprotected scenarios from the past month.

I think it's safer to take scenarios from the past 3 months or so if that's possible performance wise.

@antw Do you know if the unsaved scenarios that you migrated manually a couple of weeks ago (e.g. https://github.com/quintel/etengine/blob/master/db/migrate/20190226130006_update_housing_stock_and_insulation_inputs_om.rb) have been marked as protected now? If not these migrations won't apply to them

redekok:
@antw Thanks! I've implemented your suggested simplifications. Are we ready to merge the PR now? And execute the migrations on beta?

db/migrate/20190320145911_change_solar_park_inputs_to_capacity.rb
antw: > @antw Do you know if the unsaved scenarios that you migrated manually a couple of weeks ago have been marked as protected now?

They are indeed now protected.

antw:
> Are we ready to merge the PR now? And execute the migrations on beta? If so, we should merge quintel/etsource#1961 first, right?

Indeed the ETSource pull request should be merged and imported into ETEngine first. Then we should be good-to-go.

PR #1057, opened 2019-06-25
Housing stock sliders

redekok:
Related to quintel/etmodel#3024:

- Migration script for changing the housing stock inputs from absolute values to percentages

https://github.com/quintel/etsource/pull/2063/ should be merged first!

### DIFF
--- a/db/migrate/20190624084227_change_housing_stock_inputs_to_percentage.rb
+++ b/db/migrate/20190624084227_change_housing_stock_inputs_to_percentage.rb
@@ -1,0 +1,153 @@
+# frozen_string_literal: true
+
+# Updates scenarios to change number_of_ households inputs to be instead
+# describe the share of each household type.
+class ChangeHousingStockInputsToPercentage < ActiveRecord::Migration[5.1]
+  INPUTS = %w[
+    apartments
+    corner_houses
+    detached_houses
+    semi_detached_houses
+    terraced_houses
+  ].freeze
+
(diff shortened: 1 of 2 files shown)


db/migrate/20190624084227_change_housing_stock_inputs_to_percentage.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    say "#{total}/#{total} (#{changed} migrated)"
+  end
+
+  # Changes an input in a `scenario` from the `from` key to the `to` key,
+  # which is the percentage of the housing type compared to the total number of residences
+  #
+  # If the `from` key is not present, returns a falsey value.
+  def update_input(scenario, from, to)
+    number_of_residences = INPUTS.sum do |key|
```
antw: We could move the number_of_residences calculation up to `update_scenarios`, below: `scenarios.find_each.with_index` and pass in the value to `update_scenario` to avoid having to calculate it five times per scenario. Update input would need to take an extra parameter so that it could be given the value:

```ruby
def update_input(scenario, from, to, number_of_residences)
  # ...
end
```

redekok:
Thanks for your suggestions, @antw! I think I've implemented all of them. Could you recheck the script? :)

antw:
I realised that it'll need a few more changes in order to be reversible. I'll go into more detail tomorrow morning.

PR #1058, opened 2019-07-01
Add Elaad curves fast charging and regular charging

DorinevanderVlies:


DorinevanderVlies:
Related to
- https://github.com/quintel/etsource/pull/2067
- https://github.com/quintel/atlas/pull/133
- https://github.com/quintel/etmodel/pull/3066

Gemfile.lock
```diff
@@ -436,4 +436,4 @@ RUBY VERSION
    ruby 2.4.2p198
 
 BUNDLED WITH
-   1.17.3
+   2.0.1
```
redekok: I think the Bundler version should be changed back from 2.0.1 to 1.17.3 because it might be causing errors otherwise?

PR #1062, opened 2019-08-02
Minor change to converter technical and financial properties table

redekok:
Fixes https://github.com/quintel/etmodel/issues/3100:

image

I've decided not to include the total installed capacity in this table, since the technical and financial properties are values for one typical unit. In addition, I wasn't sure if it was technically feasible in ETEngine.


redekok:
> Perhaps we should also do the same for the other capacity attributes (on lines 68, 105, 143, 193, 274, 289, 304, and 319)?

Done!

PR #1084, opened 2020-01-20
Interleaved heat network

antw:
This PR changes the heat network calculation from being calculated after Fever+Merit, to being *between* them.

### Before:

* Fever / Electricity merit order
  * Fever (hour 1) → Merit (hour 1)
  * Fever (hour 2) → Merit (hour 2)
  * ...
  * Fever (hour 8760) → Merit (hour 8760)
* **Heat network (0 → 8760)**
* Network gas (0 → 8760)
* Hydrogen (0 → 8760)

### After:

* Fever / **Heat network** / Electricity merit order
  * Fever (hour 1) → **Heat network (hour 1)** → Merit (hour 1)
  * Fever (hour 2) → **Heat network (hour 2)** → Merit (hour 2)
  * ...
  * Fever (hour 8760) → **Heat network (hour 8760)** → Merit (hour 8760)
* Network gas (0 → 8760)
* Hydrogen (0 → 8760)

This approach is more complex and requires a custom implementation of "self: ..." curves just for the heat network. However, as an upside, we can now better represent those dispatchable heat producers which create electricity loads based on their heat production.

In order to respect the desire to calculate heat from April to March, instead of January to December, both Fever and the electricity merit order are now also calculated from April to March.

Where we have electricity producers which create heat as a side effect, we cannot provide that heat to the heat network in the same hour as the electricity is produced. This is because by the time we calculate the electricity production, we have already calculated the heat loads for the hour. As a result, this heat will be available one hour later. These nodes will therefore have no heat production in the first hour of the year.

```ruby
EACH(
  V(energy_chp_ultra_supercritical_coal, electricity_output_curve)[2190...2194],
  V(energy_chp_ultra_supercritical_coal, steam_hot_water_output_curve)[2190...2194]
)

# [
#   [
#     1,216.9125000000001,
#     1,216.9125000000001,
#     1,216.9125000000001,
#     1,216.9125000000001,
#   ],
#   [
#     0.0,
#     456.34218749999997,
#     456.34218749999997,
#     456.34218749999997,
#   ],
# ]
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,7 +51,7 @@ gem 'parallel'
 
 # own gems
 gem 'rubel',         ref: 'e36554a',  github: 'quintel/rubel'
-gem 'quintel_merit', ref: '6361a0f',  github: 'quintel/merit'
+gem 'quintel_merit', ref: '8f1e878',  github: 'quintel/merit'
 gem 'fever',         ref: 'f80677d',  github: 'quintel/fever'
 gem 'turbine-graph', '>=0.1',         require: 'turbine'
 gem 'refinery',      ref: '316c8b1',  github: 'quintel/refinery'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -24,8 +24,8 @@ GIT
 
 GIT
   remote: https://github.com/quintel/merit.git
-  revision: 6361a0fdb06e9e33f7229768c9f6789c464c33d0
-  ref: 6361a0f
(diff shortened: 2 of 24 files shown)


antw:
You may disregard the Rubocop/Hound messages. It is wrongly flagging some code for not being consistent with the style guide (when in fact it's fine) and applying other rules (like method/block length) that it shouldn't be.

PR #1085, opened 2020-01-21
Support decay factor in MeritFacade::StorageAdapter

antw:
This PR updates the `storage.decay` attribute to be treated as a factor by which the stored energy will decay each hour. For example:

* `storage.decay = 0.0` (or attribute not set) – no energy decays
* `storage.decay = 0.5` – 50% of the energy stored in the buffer decays each hour (100 stored becomes 50, then 25, etc)

In the energy_heat_network_storage node, the decayed energy will be part of the dumped energy.

Note that this also affects two nodes in the electricity merit order; these will have to be updated prior to merging this.

* [households_flexibility_p2p_electricity](https://github.com/quintel/etsource/blob/889de56e214ff61587744d5bcdfa5cecd33cb7ed/nodes/households/households_flexibility_p2p_electricity.converter.ad#L12)
* [transport_car_using_electricity](https://github.com/quintel/etsource/blob/889de56e214ff61587744d5bcdfa5cecd33cb7ed/nodes/transport/transport_car_using_electricity.converter.ad#L10)

### DIFF
--- a/app/models/qernel/merit_facade/storage_adapter.rb
+++ b/app/models/qernel/merit_facade/storage_adapter.rb
@@ -45,18 +45,30 @@ module Qernel
       def producer_attributes
         attrs = super
 
-        unless infinite_storage?
-          attrs[:reserve_class] = Merit::Flex::SimpleReserve
-        end
-
         attrs[:input_capacity_per_unit] =
           source_api.input_capacity ||
           source_api.output_capacity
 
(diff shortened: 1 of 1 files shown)


ChaelKruip:
> Note that this also affects two nodes in the electricity merit order; these will have to be updated prior to merging this.

@michieldenhaan @marliekeverweij are you able to pick this up?

PR #1093, opened 2020-02-28
Add 3 decimal point formatter for traditional_heat presenter group

michieldenhaan:
Rounds heating capacity to 3 decimal places to ensures that capacity is not rounded to zero for heating technologies

See: https://github.com/quintel/etmodel/issues/3241


noracato:
It can be confusing for dutch users to see three decimal points, for example 5.000MW in agriculture crude oil. We can either leave it like it is, with one decimal point, or try two decimal points.

michieldenhaan:
Fair point, I think we should leave this for now as there's seems to be no easy fix. Flexible units is probably the way to go as suggested here: https://github.com/quintel/etmodel/issues/3233

PR #1106, opened 2020-04-28
Merit storage may be share of another node's mean

antw:
Allows setting storage output capacity to be the mean hourly load of another node.

Merge order:

1. quintel/etsource#2268
2. This PR second
3. quintel/etmodel#3331

### DIFF
--- a/app/models/qernel/merit_facade/storage_adapter.rb
+++ b/app/models/qernel/merit_facade/storage_adapter.rb
@@ -15,6 +15,12 @@ module Qernel
 
         inject_infinite! if infinite_storage?
 
+        # If the capacity is set dynamically by this adapter, reflect the change
+        # on the source node.
+        if @config.output_capacity_from_demand_of
+          inject_dynamic_output_capacity!
+        end
+
         # If the output slot has two links, one a share link and one inversed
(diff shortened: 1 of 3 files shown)


app/models/qernel/merit_facade/storage_adapter.rb
```diff
@@ -98,6 +108,27 @@ def inject_infinite!
...
+          target_api.heat_output_capacity = capacity
+        end
+      end
+
+      # Internal: When "output_capacity_from_demand_of" is set, set the capacity
+      # of the storage to be equal to the average demand of another node,
+      # multiplied by a constant chosen by the user.
+      def capacity_from_other_demand
+        node = @context.graph.converter(@config.output_capacity_from_demand_of)
+        avg_load = node.demand / 8760
```
michieldenhaan, line 128: Shouldn't this also divide by 3600 to convert MJ to MW(h)?

antw:
> For some reason storage behaviour is odd when testing this on my local machine.

Sorry! I spent a while head-scratching over this issue earlier, but then somehow pushed the wrong commit to GitHub. It should be fixed it the latest version.

PR #1127, opened 2020-06-18
Fix incorrect P2H FLH and demand calculation

### LINKED ISSUE #1126: Heat produced by P2H in industry should not get lost in hybrid_heat - nodes (on BETA & LIVE)
@ChaelKruip and I were looking into [this scenario](https://pro.energytransitionmodel.com/saved_scenarios/9345) and found that something strange is going on in the hybrid-heat nodes:

For example in the [industry_chemicals_other_hydrogen_hybrid_heat](https://engine.energytransitionmodel.com/inspect/765039/converters/industry_chemicals_other_hydrogen_hybrid_heat) the heat produced by power to heat is reduced if I install 11890 MW P2H:
Screenshot 2020-06-18 at 14 40 44

It does not happen when I reduce the p2h capacity (for example to 5615 MW), probably then the installed P2H output capacity is lower than the heat demand capacity of chemical other:
Screenshot 2020-06-18 at 14 42 32

We maybe think this issue arose after the quick P2H in industry "fix" we did last week (which is not really a fix actually) (issue https://github.com/quintel/etengine/issues/1118)

Related to this issue is that before the deploy the industry_chemicals_other_burner_hydrogen node in this scenario was producing heat, but after the deploy the heat output is set to zero. Maybe we need to create a separate issue for this.

antw:
Fixes incorrect calculation of full load hours and demand for P2H, when input capacity is limited by total demand.

The calculation of FLH should be based on the _real_ input capacity of the node, not the artificially-limited capacity of the merit participant.

Screenshot 2020-06-18 at 14 14 11

Closes #1126

### DIFF
--- a/app/models/qernel/merit_facade/power_to_heat_adapter.rb
+++ b/app/models/qernel/merit_facade/power_to_heat_adapter.rb
@@ -12,8 +12,10 @@ module Qernel
       def inject!
         super
 
+        production = participant.production
+
         full_load_hours =
-          participant.production / input_efficiency / (
+          production / input_efficiency / (
             source_api.input_capacity *
             participant.number_of_units *
             3600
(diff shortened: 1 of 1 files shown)


marliekeverweij:
Tested it and works perfectly:
Screenshot 2020-06-18 at 15 42 29

PR #1131, opened 2020-08-27
Fix incorrect demand on flex with zero output capacity

antw:
Electricity flex technologies with no output capacity would have a demand of zero set in the graph, rather than their real electricity demand. This is because four months ago we added support for heat storage, and correct calculation of its demand needed to be based on its output capacity. I wrongly opted all flexibilities technologies into this behaviour; technologies with zero electricity output capacity (P2H, P2H) would therefore receive a demand of zero in the graph.

This, in turn, would cause incorrect calculation of CO2 when using these technologies and failing Mechnical Turk tests.

Their demands are now set based on their input capacity – since all flex will have input, but not all have output – and a custom adapter has been added for heat storage.

Please test or merge simultaneously with https://github.com/quintel/etsource/pull/2340.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem 'parallel'
 
 # own gems
 gem 'rubel',         ref: 'e36554a',  github: 'quintel/rubel'
-gem 'quintel_merit', ref: 'a0bc994',  github: 'quintel/merit'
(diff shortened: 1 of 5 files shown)


redekok (approved):
LGTM! I've checked the demands against the curve calculated by Merit for nearly all P2G and P2H technologies. Numbers seem to be fine.

PR #1132, opened 2020-09-29
Price-sensitives receive equal shares of excess/dispatchables

### LINKED ISSUE #1129: Electricity interconnectors with the same hourly price should distribute electricity equally
As mentioned by @ChaelKruip during the interconnectors wrap-up this week: electricity interconnectors with the same hourly price should distribute electricity equally. At the moment the interconnector with a lower number imports/exports electricity first until the max capacity is reached. 

For example interconnector 1 and interconnector 2 both have 35 euro/MWh. At the moment interconnector 1 is first used for import/export and than interconnector 2. However, actually both interconnectors should import/export the same amount of electricity in that hour.

@antw , how much time do you think this will take to implement?

antw:
* Price-sensitive users now receive equal shares of excess.
* Price-sensitive users now receive equal shares of available dispatchable energy.

The performance penalty is hard to quantify. On my laptop, three interconnectors with the dispatchables toggle turned on, compared with one interconnector and the toggle off, costs maybe 100ms. I think these changes are responsible for roughly half of that. I expect you'll want to merge this, but I feel like I have to reiterate that our use of Ruby for hourly calculations is becoming more ridiculous with every new feature.

P.S. My brain is broken from writing `while` loops in Ruby.

4golgu

## Excess energy

| One interconnector | Two interconnectors |
| :---: | :---: |
| equal-excess-1 | equal-excess-2 |

## Dispatchables toggle on

| Two interconnectors | Three interconnectors |
| :---: | :---: |
| Screenshot 2020-09-29 at 15 57 00 | Screenshot 2020-09-29 at 15 57 09 |

Closes #1129

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem 'parallel'
 
 # own gems
 gem 'rubel',         ref: 'e36554a',  github: 'quintel/rubel'
-gem 'quintel_merit', ref: '6a614e3',  github: 'quintel/merit'
+gem 'quintel_merit', ref: '74b140d',  github: 'quintel/merit'
 gem 'fever',         ref: 'f80677d',  github: 'quintel/fever'
 gem 'turbine-graph', '>=0.1',         require: 'turbine'
 gem 'refinery',      ref: '72eacf8',  github: 'quintel/refinery'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -24,8 +24,8 @@ GIT
 
 GIT
   remote: https://github.com/quintel/merit.git
-  revision: 6a614e3ed3c2d3c4d1a324c29d1829829d937d16
-  ref: 6a614e3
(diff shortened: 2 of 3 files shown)


marliekeverweij (approved):

marliekeverweij:
> P.S. My brain is broken from writing while loops in Ruby.

😩

marliekeverweij:
> P.S. My brain is broken from writing while loops in Ruby.

😅

>  I expect you'll want to merge this, but I feel like I have to reiterate that our use of Ruby for hourly calculations is becoming more ridiculous with every new feature.

Sounds wise to have a good look at our use of Ruby for hourly calculations. But maybe not now since your brain is broken and you're maybe to deep in it now to have a good look. 

I would propose to merge this and add a new to-do to the developer sprint for reiterating the hourly calculations. What do you think @ChaelKruip @antw ?

ChaelKruip (approved):
LGTM!

PR #1135, opened 2020-11-03
Reset molecule graph between calculations

antw:
When the energy graph is reset (after Causality) the molecule graph should also be reset, ensuring that values from the re-calculated molecule graph are propagated.

Basecamp: [Merit impact does not propagate through the molecule graph?](https://3.basecamp.com/3797409/buckets/16508470/todos/3174723949)


michieldenhaan (approved):
Thanks!

PR #1138, opened 2020-11-12
CCUS migrations

michieldenhaan:
- Set (new) 'other utilisation' input to zero for all scenarios.
- Fix multiplication error in previous energy CCS migration
- Add temporary migration for BETA scenarios


antw (approved):
Looks good to me! :+1:

PR #1139, opened 2020-11-18
Add migration for households P2P

michieldenhaan:
Specs of households P2P have been updated (quintel/etsource@4fe7e510231a9d9d8b413e4d73f73c27bd450bd7). The default volume of 1 P2P unit is now 50% lower than before. To limit the impact on scenarios, the households P2P slider value is doubled to ensure the available storage volume remains constant. For scenarios with an initial slider value of >50%, doubling the slider value is not fully possible as the max slider value is 100%.

See: https://github.com/quintel/etsource/pull/2376


antw (approved):
Looks good to me!

PR #1140, opened 2020-11-27
Add migration to fix district heating cost sliders

michieldenhaan:
Scenarios based on the featured Climate Agreement scenario wrongfully assume a (non-sensical) 99.2% decrease in district heating costs. This should be a 20% decrease according to the scenario authors. The input value was (somehow?) changed from 80 to 0.8, a factor 100 difference). This migration looks up all scenarios with the faulty 0.8 values and overwrites them with the correct 80 value.

### DIFF
--- a/db/migrate/20201126215921_fix_heat_cost.rb
+++ b/db/migrate/20201126215921_fix_heat_cost.rb
@@ -1,0 +1,19 @@
+require 'etengine/scenario_migration'
+
+class FixHeatCost < ActiveRecord::Migration[5.2]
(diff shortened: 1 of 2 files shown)


antw (approved):
:+1:

antw:
@michieldenhaan Does this also need to go to production now?

michieldenhaan:
I think this can wait until next deploy. I've already corrected the featured scenario manually, so new scenarios based on the preset will have the correct settings.

PR #1149, opened 2021-01-19
Update node properties table

marliekeverweij:
This update includes:
- All technologies (except for heating technologies built environment and solar production): removed typical capacity per unit and added total installed capacity
- All technologies except for heating technologies built environment: presented costs as costs / MW and not per unit
- Storage P2P technologies: removed typical storage volume per unit and added total storage volume
- Vehicle-to-grid: add presenter_group v2g so that round trip efficiency can be shown
- All technologies except for solar panels: removed land use per unit, as very often zero
- All technologies: removed construction time, added decommissioning costs and installations costs with investment costs
- Additional costs for CCS: only presented these costs for relevant technologies (few power plants, hydrogen)

Goes with:
https://github.com/quintel/etsource/pull/2400
https://github.com/quintel/etmodel/pull/3548

Issues that @antw will fix before merging:

- [x] scaling of m2 / km2
- [x] scaling of EUR
- [x] T / hour -> T CO2 / hour (if possible!)
- [x] solar pv panels do not show their capacity at the moment

### DIFF
--- a/app/serializers/node_serializer_data.rb
+++ b/app/serializers/node_serializer_data.rb
@@ -591,8 +591,14 @@ module NodeSerializerData
   CO2_CAPTURE_ATTRIBUTES_AND_METHODS =
     {
       technical: {
+        'typical_input_capacity * 8760 * number_of_units' => {
+          label: 'Total installed capture capacity',
+          key: :total_installed_capture_capacity,
+          unit: 'T_CO2 / year',
+          formatter: FORMAT_KILO
+        },
         'typical_input_capacity * 8760' => {
-          label: 'Yearly capture capacity',
+          label: 'Capture capacity per unit',
           key: :typical_input_capacity_co2_capture,
           unit: 'T_CO2 / year',
           formatter: FORMAT_KILO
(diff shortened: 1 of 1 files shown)


marliekeverweij:
@antw could you please squash my commits? :)

antw (approved):
:+1:

app/serializers/node_serializer_data.rb
```diff
@@ -471,12 +645,17 @@ module NodeSerializerData
       }
     },
     cost: {
-      'total_initial_investment_per(:plant)' => { label: 'Initial purchase price', unit: 'EUR' },
-      'cost_of_installing_per(:plant)' => { label: 'Cost of installing', unit: 'EUR' },
-      'decommissioning_costs_per(:plant)' => { label: 'Decommissioning costs', unit: 'EUR' },
-      'fixed_operation_and_maintenance_costs_per(:plant)' => {
+      'total_initial_investment_per(:mw_typical_input_capacity) + cost_of_installing_per(:mw_typical_input_capacity) + decommissioning_costs_per(:mw_typical_input_capacity)' => {
```
michieldenhaan, line 648: `Total_initial_investment` already includes cost of installing. I think we can use `total_investment_over_lifetime` to capture initial investment, installing and decomissioning.

app/serializers/node_serializer_data.rb
```diff
@@ -457,11 +549,93 @@ module NodeSerializerData
...
+        key: :round_trip_efficiency,
+        unit: '%',
+        formatter: FORMAT_FAC_TO_PERCENT
+      }
+    }
+  }.freeze
+
+  CO2_GENERIC_ATTRIBUTES_AND_METHODS = {
+    cost: {
+      'total_initial_investment_per(:plant)' => {
```
michieldenhaan, line 567: Maybe I misinterpret this but I thought we wanted to show CCUS costs per capacity (tonne per year / hour?) rather than per unit?

michieldenhaan (changes requested):

app/serializers/node_serializer_data.rb
marliekeverweij: Yes that was indeed my intention, but doesn't work at the moment. I'll write it down on the future improvement list for this table :)

app/serializers/node_serializer_data.rb
marliekeverweij: Fixed this, good idea!

michieldenhaan (approved):

PR #1156, opened 2021-02-19
Support esdl upload

### LINKED ISSUE #1152: Add support for ESDL files in ScenarioAttachment
Part of quintel/etmodel#3535

noracato:
Adds ESDL files as possible `ScenarioAttachments`, and adds an api to upload and download them.
Also returns information about if the scenario was started from ESDL in the scenario serializer, so that etmodel can act accordingly.

Closes #1152
Related to quintel/etmodel#3624

### DIFF
--- a/app/models/file_upload_handler.rb
+++ b/app/models/file_upload_handler.rb
@@ -11,7 +11,8 @@ class FileUploadHandler
   # key      - The key of the ScenarioAttachment, e.g. 'esdl_file'
   # scenario - The scenario the file should be attached to
   def initialize(file, filename, key, scenario)
-    @file = file
+    @file = Tempfile.new(filename)
+    @file.write(file)
     @filename = filename
     @key = key
     @scenario = scenario
@@ -21,11 +22,13 @@ class FileUploadHandler
   def call
     attachment = ScenarioAttachment.create!(key: @key, scenario: @scenario)
 
+    @file.rewind
     attachment.file.attach(
-      io: StringIO.new(@file),
+      io: @file,
       filename: @filename,
       content_type: 'text/xml'
     )
+    @file.unlink
   end
 
   def valid?
@@ -33,10 +36,16 @@ class FileUploadHandler
 
     if @scenario.attachments.find_by(key: @key)
       @errors.push("This scenario already has a file of type #{@key} attached.")
+      return false
     end
 
     unless ScenarioAttachment.valid_non_curve_keys.include?(@key)
       @errors.push("This handler cannot attach files of type #{@key}.")
+    end
+
+    if @key == 'esdl_file'
+      header = File.open(@file.path) { |file| file.gets(nil, 1024) }
+      errors.push('This file does not contain ESDL') unless header.include?('<esdl:EnergySystem')
     end
 
     @errors.none?
--- a/app/models/file_upload_handler.rb
+++ b/app/models/file_upload_handler.rb
@@ -10,10 +10,8 @@ class FileUploadHandler
   # filename - The name of the file
   # key      - The key of the ScenarioAttachment, e.g. 'esdl_file'
   # scenario - The scenario the file should be attached to
-  def initialize(file, filename, key, scenario)
-    @file = Tempfile.new(filename)
-    @file.write(file)
-    @filename = filename
+  def initialize(file, key, scenario)
+    @file = file
     @key = key
     @scenario = scenario
   end
@@ -22,13 +20,11 @@ class FileUploadHandler
   def call
     attachment = ScenarioAttachment.create!(key: @key, scenario: @scenario)
 
-    @file.rewind
     attachment.file.attach(
       io: @file,
-      filename: @filename,
+      filename: @file.original_filename,
       content_type: 'text/xml'
     )
-    @file.unlink
   end
 
   def valid?
--- a/app/models/file_upload_handler.rb
+++ b/app/models/file_upload_handler.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+# Handles the upload of files that aren't curves, or don't need any (pre)processing like esdl files
+class FileUploadHandler
+  attr_reader :errors
+
+  # Public: creates a new handler
+  #
+  # file     - A String version of the file that should be attached
+  # filename - The name of the file
+  # key      - The key of the ScenarioAttachment, e.g. 'esdl_file'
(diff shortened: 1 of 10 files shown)


antw (changes requested):
This looks good to me, but I think I'd like to see some tests of the upload handler, and some request specs for the file upload/download.

I expect the existing specs for curves should be fairly easy to adapt.

* [Fetching curve data](https://github.com/quintel/etengine/blob/master/spec/requests/api/v3/custom_curve_spec.rb#L59)
* [Uploading a valid curve](https://github.com/quintel/etengine/blob/master/spec/requests/api/v3/custom_curve_spec.rb#L96)

Would you have time to do this?

Otherwise, this looks really good to me.

noracato:
Very valid points! I'll create them now, should be finished tomorrow morning :)

spec/models/file_upload_handler_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+            .reload
+            .attachments
+            .find_by(key: 'esdl_file')
+            .present?
+        }
+        .from(false).to(true)
+    end
+  end
+
+  # We don't really care - should we?
```
noracato, line 41: @antw do you think we should ensure the file is ESDL/xml before we allow it to be attached? Only the etm-esdl app uses this feature right now, so that kind of ensures the file being a valid ESDL file. However, the API is open to everyone that wants to use it.

antw: Sure, we could do.

One naive approach is just to read the first 1kb or so, and check that there's something that looks like an `edsl:EnergySystem` tag:

```
header = File.open(path_to_uploaded_tempfile) { |file| file.gets(nil, 1024) }

unless header.include?('<esdl:EnergySystem')
  errors.add('...')
end
```

It's not exactly foolproof (the file could still not be valid ESDL), but it might help catch some user errors.

If you wanted, we could do more detailed validations by loading the XML file with Nokogiri and checking for the existence of certain tags or attributes, but I'm not sure if you think that's worth it?

I feel as though checking for the esdl:EnergySystem tag in the first 1k is probably good enough.

spec/requests/api/v3/esdl_files_controller_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+            .present?
+        }
+        .from(false).to(true)
+    end
+  end
+
+  context('when uploading an esdl file larger than 5MB') do
+    let(:request) do
+      put url, params: {
+        file: fixture_file_upload('files/huge_invalid_esdl_file.esdl', 'text/xml').read,
```
antw, line 114: Could we [generate this file dynamically](https://github.com/quintel/etengine/blob/05e87d469e2d2e2011231e17ceadc6508a30757d/spec/requests/api/v3/custom_curve_spec.rb#L230-L240) in the test, rather than storing a 5MB file in the repo, or would that defeat the point of the test? I guess this is harder to do for ESDL than it is for a simple curve file.

spec/models/file_upload_handler_spec.rb
noracato: > I feel as though checking for the esdl:EnergySystem tag in the first 1k is probably good enough.

Agreed! The fact that it is an ESDL file is more important than it being valid ESDL, in my opinion. We would have to keep up with new ESDL versions coming out every few months, doing that in just the etm-esdl app is hard enough. 

Thanks for your code snippet!

spec/requests/api/v3/esdl_files_controller_spec.rb
noracato: It shouldn't be so hard if we don't check for validity. We only need the xml and ESDL headers and closing brackets. Then we can just fill it up with an endless amount of wind turbine assets.

noracato: There's now a dynamically generated file containing 25,000 wind turbines!

app/models/file_upload_handler.rb
```diff
@@ -11,7 +11,8 @@ class FileUploadHandler
   # key      - The key of the ScenarioAttachment, e.g. 'esdl_file'
   # scenario - The scenario the file should be attached to
   def initialize(file, filename, key, scenario)
-    @file = file
+    @file = Tempfile.new(filename)
```
antw, line 14: I'm unclear why the uploaded contents are written to a tempfile. Was this because of my suggestion for scanning the first 1k of ESDL files?

I mistakenly thought the `file` was already a tempfile, but now I see it's a string. Could this be reverted with the validation doing...

```ruby
errors.add('...') unless @file[0...1024].include?('<esdl:EnergySystem')
```

... or am I missing something?

---

I was tripped by by `file` being a string, because I was expecting an [ActionDispatch::Http::UploadedFile](https://api.rubyonrails.org/v6.0.3.3/classes/ActionDispatch/Http/UploadedFile.html), which is what would normally result from an uploaded file. Instead it appears as though the file contents are provided to ETEngine as a string rather than using multipart/form-data. I'm a bit split on this:

* It's not consistent with curve uploads which use multipart/form-data, and therefore may be confusing to API users.
* I'm unclear whether there could be encoding issues providing the file as a string rather than a binary upload. I'm guessing not, but... 🤷‍♂️
* This would be an issue if you ever wanted someone to upload an image, or some other form of binary/encoded data.
* It wouldn't be possible to create an HTML form to add a file directly; we'd need a proxy which received the upload, read the contents into a string, then submitted it to ETEngine. Though to be fair, it's not clear if we'd ever need this.

Would this be a real pain to change in the etm-esdl app? If so, I'm fine leaving it for now, but it feels awkward to design the ETEngine API around etm-esdl, rather than the other way around.

Sorry to keep coming back to you with more. 😟

noracato: No worries, that's how you learn 😄 I'm rushing this a bit, because I was hoping it could join this deploy, but maybe I need to slow down because I'm starting to get sloppy... Sorry about that!

Yep, I saw your earlier suggestion and I thought: why did I ever choose to make this a String upload? So I already went back to the etm-esdl app and tried to change the API to send form-data to etengine, because I agree that all the upload API's should be similar. However, the ESDL file is changed by the app before it gets sent to ETEngine and after spending 20min trying to convince Flask to put it back into form-data, I thought 🤷 So here we see the result of that haha.

So again - excuse my sloppiness. I'll slow down and go make it pretty!

antw (approved):

PR #1161, opened 2021-03-22
Add units to the headers of all csv exports

### LINKED ISSUE #1160: Add units to data export tables
On a regular basis I get clarification requests for the data export tables.
I think it would be a good improvement if we add units to the data export tables.

Either in the names of the attributes or in an extra row below the column titles.

noracato:
Closes #1160

Units added:

- MW for hourly curves (except price)
- MJ for energy_flow
- kg for molecule_flow
- MJ for application demand
- A variety of units for production_parameters

### DIFF
--- a/app/controllers/api/v3/export_controller.rb
+++ b/app/controllers/api/v3/export_controller.rb
@@ -24,14 +24,17 @@ module Api
       #
       # Returns a CSV file containing the energetic inputs and outputs of every node in the graph.
       def energy_flow
-        send_csv(NodeFlowSerializer.new(scenario.gql.future.graph), 'energy_flow.%d.csv')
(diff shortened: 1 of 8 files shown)


antw (approved):
Looks good to me; thanks @noracato!

michieldenhaan:
Nice work!

I think we should communicate this improvement clearly for next deploy as this change might 'break' some users' workflows (e.g. if they use Excel or some other tool that relies on matching column names). (cc: @marliekeverweij )

marliekeverweij:
> I think we should communicate this improvement clearly for next deploy as this change might 'break' some users' workflows (e.g. if they use Excel or some other tool that relies on matching column names). (cc: @marliekeverweij )

Good point!

PR #1179, opened 2021-06-03
Ensure electrolyser shares are consistent with flows

### LINKED ISSUE #1174: Inconsistent demand and output of wind electrolysis
Scenario: 835873 (PRO)

Graph shows an input of 81 PJ, demand of 76 PJ and hydrogen output of 53.5 PJ. So 5 PJ disappears from input to demand. Also output efficiency is 66%. 76 * 0.66 = 50.2 PJ hydrogen output, not 53.5 PJ. Querying `V(energy_hydrogen_electrolysis_wind_electricity,output_of_hydrogen)` also returns 50.2 PJ hydrogen output but graph uses 53.5 PJ.

image

antw:
Sets electrolyser shares prior to recalculation of the graph. 2d5733f required that Reconciliation be moved to after recalculation of the graph so that demand changes made by Merit/Fever were reflected in and hydrogen.

This had the effect of setting electrolyser shares too late: the graph would already have been calculated, and flows were therefore not consistent with the shares set by ElectrolyserAdapter.

As electrolyser demand is known prior to Merit, and cannot be changed (due to them having dedicated electricity sources which are excluded from Merit), we can set their demands prior to recalculation of the graph.

A `before_graph_recalculation` method is added to Reconciliation which is triggered by Causality.

Closes #1174

### DIFF
--- a/app/models/qernel/causality/reconciliation_wrapper.rb
+++ b/app/models/qernel/causality/reconciliation_wrapper.rb
@@ -29,6 +29,15 @@ module Qernel
         end
       end
 
+      # Internal: A hook which is called by Causality prior to the recalculation of the energy
+      # graph.
+      #
+      # This is needed to allow electrolyzers to set the shares of some edges prior to the graph
+      # being recalculated.
+      def before_graph_recalculation
+        @managers.each(&:before_graph_recalculation)
+      end
+
       private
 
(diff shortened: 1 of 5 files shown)


antw:
Screenshot 2021-06-03 at 12 51 17

michieldenhaan (approved):

PR #1180, opened 2021-06-03
Fix recursive factor with efficiencies exceeding 1

### LINKED ISSUE #1172: Wrong primary demand with efficiencies > 100%?
Scenario on BETA: 1435024. Very simple region with only Household demand (other sectors are 0).

The `energy_production_wood_pellets` node has an output of wood pellets (= primary demand) of 1.55 PJ.
The `primary_demand_of_wood_pellets` of `households_final_demand_steam_hot_water`, `households_final_demand_electricity` and `energy_heat_dumped_steam_hot_water` combined is 1.73 PJ, more than the total primary production of wood pellets.

The culprit seems to be `energy_chp_local_wood_pellets`. This node has a combined heat and electricity efficiency of 111%. When setting CHP capacity to zero and supplying heat with a wood pellet heater instead (total efficiency 90%) the mismatch disappears.

My guess of what is going on here is that the individual carrier efficiencies of the CHP are both < 100% (heat efficiency 82%, electricity efficiency 29%). However, the combined efficiency is > 100%. Suppose wood pellet input is 100 MJ. Then heat output is 82 MJ and electricity output is 29 MJ. The primary demand of the heat output should be 74 MJ, the primary demand of electricity output 26 MJ because 74% of output is heat and 26% of output is electricity. However, because both individual outputs are < 100%, my guess is that 82 MJ primary demand is assigned to heat and 29 MJ to electricity, leading to an overestimate of total primary demand.

@antw Do you know if this is indeed the case? And if this could be fixed with the current implementation of the primary demand method?

Notifying @ChaelKruip, @marliekeverweij 
Revisit of https://github.com/quintel/etengine/issues/936.

antw:
When the outputs of a node sum to a value greater than 1, recursive factor values are exaggerated. For example, if we have a graph like where a node outputs energy to two suppliers in different carriers, with a total output greater than 1, recursive factor respects the output conversions of the producer and calculates an incorrect primary demand for the consumers:

1-current

The expected behaviour is that each consumer should have 50 *primary* demand: although they each have 60 *demand*, they require 50 *primary demand* on the producer node to generate that 60.

2-expected

This commit adds an `output_efficiency_compensation_factor` which is used by `recursive_factor` (but not `recursive_factor_without_losses`) to compensate for nodes with output shares exceeding 1.

The "without losses" does not have this compensation, as it is typically used to calculate shares (`sustainability_share`).

Fixes #1172

---

```ruby
EACH(
  SUM(V(
    households_final_demand_steam_hot_water,
    households_final_demand_electricity,
    energy_heat_dumped_steam_hot_water,
    primary_demand_of_wood_pellets
  )),
  V(energy_production_wood_pellets, primary_demand_of_wood_pellets)
)
```

### Before

```
[
  1,724,900,051.4137213,
  1,553,964,010.2826319,
]
```

### After

```
[
  1,553,964,010.2826319,
  1,553,964,010.2826319,
]
```

### DIFF
--- a/app/models/qernel/recursive_factor/base.rb
+++ b/app/models/qernel/recursive_factor/base.rb
@@ -69,7 +69,7 @@ module Qernel::RecursiveFactor::Base
         # "compensation factor" to include losses.
         #
         # See https://github.com/quintel/etengine/issues/518.
-        loss_compensation_factor = parent.query.loss_compensation_factor
+        parent_output_compensation_factor = parent.query.output_compensation_factor
 
         # What part is considered to be contributing to the outcome?
         # (e.g. 80% when free_co2_factor is 20%). This is 100% when the
@@ -79,7 +79,7 @@ module Qernel::RecursiveFactor::Base
         if demanding_share.zero?
           # The node has no demand, we can safely omit it.
           0.0
-        elsif loss_compensation_factor.zero?
+        elsif parent_output_compensation_factor.zero?
           # The node is 100% loss, so there is no point in recursing
           # further on this edge.
           0.0
@@ -93,7 +93,7 @@ module Qernel::RecursiveFactor::Base
             include_abroad: include_abroad
           )
 
-          demanding_share * loss_compensation_factor *
+          demanding_share * parent_output_compensation_factor *
             node_share * parent_value
         end
       end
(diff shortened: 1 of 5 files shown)


antw:
@michieldenhaan I think this is ready now.

This sent me down a rabbit hole investigating recursive factor. I overhauled many of the tests and discovered other issues, mainly with `sustainability_share` and bio-emissions with inputs or outputs that don't sum to 1.0.

The improved tests and bio-emissions fixes are already on `master`. `sustainability_share` changes are pending while I figure out the best approach. This PR specifically addresses when the sum of outputs on nodes exceed 1.0.

michieldenhaan (approved):

PR #1182, opened 2021-06-08
Fix mismatch between heat curves and output

### LINKED ISSUE #1175: Small mismatch in primary demand when using CHPs and Merit
Scenario: 1435329 (BETA)

I am testing whether:

```
SUM(V(G(primary_energy_demand),primary_demand)) - 
SUM(V(G(final_demand_group),G(energy_export), energy_flexibility_curtailment_electricity, industry_locally_available_refinery_gas_for_chemical, primary_demand))
```
is zero. I.e. total primary demand should either be traceable to final demand or export. Curtailment and refinery gas are also included in the query because these nodes currently are 'dead ends' in the graph but are not considered final demand or export.

For the present (NL2015) the above query does not return zero because of the biomass CHP with an efficiency of 111% (see: https://github.com/quintel/etengine/issues/1172).

In 1435329 I have turned this CHP off and tried to decrease all energy carriers except electricity, gas and heat as much as possible to simplify the analysis. When running the query, I get a result of 27,196,895 MJ. So 27 TJ primary demand is not account for by final demand, export, curtailment and refinery gas. My suspicion is that this is caused by CHPs, Merit and possibly the interaction between Merit and time-resolved heat: 
- When turning off the "[Large-scale gas plant for district heat (CHP)](https://beta-pro.energytransitionmodel.com/scenario/supply/electricity/gas-plants)" the difference drops to 0.00048828125 MJ. 
- When turning Merit off, the difference disappears completely: 0.0. 
- When swapping the gas CHP for a coal CHP (6 GW) the difference is 9,520,412 MJ.
- When switching on an industry CHP (where heat demand is not time-resolved) the different is again 0.00048828125 MJ.

Could this have something to do with CHP heat being available one hour after electricity production? Maybe heat produced in the final hour of the year 'disappears'/is not registered in the graph but the input fuel is? Resulting in a small mismatch in primary demand (input fuels) and final demand (heat demand) in the graph?

antw:
When a heat producer also participates in the electricity merit order, the heat curve does not have a load to represent the final hour in the year (all heat loads a shifted one hour into the future, so this manifests with no heat load in the _first_ element of the load curve).

This results in slightly too much heat leaving the producer in the graph since the share of heat output does not account for the hour in which none is produced.

`ProducerAdapter` has been adjusted to change the share of heat to ensure that graph flows match the heat output curve. 

## Query from #1175

```ruby
SUM(V(G(primary_energy_demand),primary_demand)) - 
  SUM(V(
    G(final_demand_group),
    G(energy_export),
    energy_flexibility_curtailment_electricity,
    industry_locally_available_refinery_gas_for_chemical,
    primary_demand
  ))
```

### Beta scenario 1435329

| Branch | Result |
| :-- | --: |
| master | `27,196,895.7 ~ ~ ~ ~ ~` |
| issues/1175-chp-outputs | `0.00024414063` |

### Beta scenario 1436077

[The simplified scenario](https://github.com/quintel/etengine/issues/1175#issuecomment-846000417) which uses coal CHPs instead of gas.

| Branch | Result |
| :-- | --: |
| master | `32,488,799.5            ` |
| issues/1175-chp-outputs | `0.0` |

### Default scenario

However, this change doesn't seem to have much impact on the mismatch in a default scenario. A mismatch remains for both the present and future.

| Branch | Present | Future |
| :-- | --: | --: |
| master | `27,634,968,521` | `27,638,966,692` |
| issues/1175-chp-outputs | `27,634,968,521` | `27,635,452,303` |

This is 1000× larger than the example scenarios. 😕 

## Heat output vs heat output curve

```ruby
EACH(
  V(energy_chp_ultra_supercritical_coal, output_of_steam_hot_water),
  PRODUCT(SUM(V(energy_chp_ultra_supercritical_coal, steam_hot_water_output_curve)), MJ_PER_MWH)
)
```

### Default scenario

To account for the missing hour, the heat output is adjusted down from 15% to 14.99828747573924%.

| Branch | Output of heat | Sum of heat curve in PJ | Diff |
| :-- | --: | --: | --: |
| master | `14,389,564,393` | `14,391,207,225` | `-1,642,832` |
| issues/1175-chp-outputs | `14,389,564,393` | `14,389,564,393` | `0` |

---

Closes #1175

### DIFF
--- a/app/models/qernel/causality/lazy_curve.rb
+++ b/app/models/qernel/causality/lazy_curve.rb
@@ -15,7 +15,7 @@ module Qernel
 
       def to_a
         if @values.last.present?
-          @values.dup
+          @values.length > length ? @values[0...length] : @values.dup
         else
           Array.new(length) { |frame| get(frame) }
         end
--- a/app/models/qernel/merit_facade/producer_adapter.rb
+++ b/app/models/qernel/merit_facade/producer_adapter.rb
@@ -21,10 +21,41 @@ module Qernel
 
       def inject!
         install_demand!
+        inject_self_shares!
         inject_curve!(:output) { @participant.load_curve }
       end
 
       private
+
+      def inject_self_shares!
+        return unless @config.subtype == :must_run
+        return unless @config.group.to_s.delete(' ') == 'self:electricity_output_curve'
+        return unless source_api.merit_order&.subtype == :dispatchable
+
+        # This heat producer is also part of the electricity merit order, where it behaves as a
+        # dispatchable. Since it is dispatchable, the electricity load for hour n is not known until
+        # the heat network is calculated for n + 1. Therefore the last hour of electricity doesn't
+        # generate any heat in the network.
+        #
+        # To ensure that the energy flows from the node are consistent with the hourly heat
+        # calculation, we must adjust the heat output.
+        #
+        # See https://github.com/quintel/etengine/issues/1175
+
+        heat_output = target_api.output(@context.carrier)
+
+        load_curve = @participant.load_curve
+        sum = load_curve.sum
+
+        # The last hour of heat is lost.
+        excess = load_curve.get(8760)
+
+        return if sum.zero? || excess.zero?
+
+        excess_share = heat_output.conversion * (excess / (sum + excess))
+
+        heat_output.conversion -= excess_share
+      end
 
       def producer_attributes
         attrs = super
--- a/app/models/qernel/recursive_factor/base.rb
+++ b/app/models/qernel/recursive_factor/base.rb
@@ -332,7 +332,10 @@ module Qernel::RecursiveFactor::Base
   #
   # Returns a numeric.
   def output_efficiency_compensation_factor
-    factor = outputs.sum { |output| output.loss? ? 0.0 : output.conversion }
+    factor = outputs.sum do |output|
+      output.carrier.loss? || output.carrier.coupling_carrier? ? 0.0 : output.conversion
+    end
+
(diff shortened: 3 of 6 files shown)


michieldenhaan:
> This is 1000× larger than the example scenarios. 😕

This seems to be introduced in https://github.com/quintel/etengine/commit/72c997135b4ca56950dc2930f8cdc423f06ebd7e. It disappears completely when turning off the steel sector.

`V(energy_distribution_coal_gas, primary_demand)` gives 87 PJ. On PRO this is 112 PJ. If we move one node to the right, `energy_steel_blastfurnace_current_transformation_cokes` has a primary demand of 98 PJ. But 100% of its production is passed onto `energy_distribution_coal_gas`, so the latter should have at least 98 PJ primary demand. Could this have something to do with the coupling carrier, causing some nodes to have > 100% outputs (though not inputs as far as I see)?

(Maybe better to move this to a separate issue, it seems to be unrelated to this PR)

michieldenhaan (approved):

antw:
> Could this have something to do with the coupling carrier

Great catch. You're right: ignoring coupling carrier causes the mismatch to fall to 483,782 (483 GJ). It's still not zero, but much better than 27 PJ.

michieldenhaan:
Thanks! This seems to be due to `energy_chp_local_engine_biogas` combined with merit being on. When switching off the biogas CHP the diff drops to 1 MJ, when increasing installed capacity the diff increases. Not sure what's so special about biogas, apart from it being the only must-run CHP we have (I think). I'll look into it some more when time permits

Issue #1202, opened 2021-10-25
Fuel production sliders should be migrated


michieldenhaan:
A few months ago fuel production time curves have been removed and [inputs](https://github.com/quintel/etsource/tree/master/inputs/supply/fuel_production) have been added for future fuel production of oil, coal, gas etc. A migration for this is missing. As a result, the present fuel production persists in the future. This typically ;eads to a higher production of fossil fuels in future years than before.

Ideally, the values of the new inputs should correspond with the original time curve values.

Please re-assign at will.

PR #1203, opened 2021-10-29
Set fuel production inputs

marliekeverweij:
Closes https://github.com/quintel/etengine/issues/1202

@noracato and I have created a scenario migration for `nl`, `nl2019`,`dk`& `de` and for 2020, 2030, 2040 & 2050, as I assume that these are the most used datasets and end years

@noracato could you check the migration? The values differ per dataset and I tried to change the script, but not sure if it works likes this


michieldenhaan (changes requested):
Two comments:
- The unit of the inputs is in PJ, I think the values in this migration are in TJ? 
- I would expect that this migration does not affect scenarios with a non-country dataset, scenarios already containing a value for the new inputs and scenarios created after the inputs have been added to the ETM. The migration log shows that 100% of the candidate scenarios are migrated. @noracato Is this because only scenarios that do not meet any of the above 3 criteria are marked as 'candidate' scenarios?

image

noracato:
> Is this because only scenarios that do not meet any of the above 3 criteria are marked as 'candidate' scenarios?

Yes, to save us some time looping through all scenarios the selected candidate scenarios are:

- Scenarios of a country (except nl2019)
- Scenarios that were created before the Fuel Production update was deployed (1 June 2021)
- Scenarios that apply to the basic 'migratable' criteria (protected, not made by mechanical turk, etc)

When looping through these candidates, each is checked to see if any of the fuel production inputs have been touched. If so, we skip this scenario. When I ran the migration on the beta database 3728/3809 scenarios were changed by the migration. That seemed like an acceptable number to me. I'm assuming you ran your migration on a production database, not sure why no one touched these sliders there yet... or do you see something else that is strange in the migration that might explain this?

michieldenhaan:
> Scenarios that apply to the basic 'migratable' criteria (protected, not made by mechanical turk, etc)

This could maybe be an issue? I guess some recent scenarios are marked as migratable but also have been created _after_ the fuel inputs have been added. In such case I would say we do not migrate, as the user had the chance to set the inputs if they wanted to.

noracato:
Don't worry. The list I put was an and-and-and condition, not or-or-or: we only take migratable scenarios that were created before the update and are based on a country.

ChaelKruip:
@noracato what is the status of this?

mabijkerk:
@noracato could you check whether this migration is still up to date and merge it before the coming Deploy (next week Tuesday)?

mabijkerk (approved):

noracato:
It still runs without any problems. But I'm not the best to judge if the results still make sense, though I wouldn't know why they won't as the migration is pretty basic.

One comment I'd like to make is that the migration is altering all elegible scenario's that have been created before the fuel update (June '21), where the fuel sliders haven't been touched, and that are based on a country that is not `nl2019`. Are we still happy with those conditions? If yes, I think we can merge 🙂

PR #1226, opened 2022-01-18
Add support for scenario metadata

noracato:
Adds a metadata field to Scenario's.

The metadata field can be set through the scenario API enpoint by a user, and can contain any data in JSON format up to 64Kb. When a new metadata object is set through the API by a user, the old metadata object is overwritten. Users can view the metadata object by setting the `detailed` parameter to `true` in their API request (similar to viewing `user_values`).

### DIFF
--- a/app/models/scenario/user_updates.rb
+++ b/app/models/scenario/user_updates.rb
@@ -100,7 +100,7 @@ module Scenario::UserUpdates
 
   # Validation method for when a user sets their metadata.
   def validate_metadata_size
-    errors.add(:metadata, 'can not exceed 64Kb') if metadata.to_s.bytesize > 64_000
+    errors.add(:metadata, 'can not exceed 64Kb') if metadata.to_s.bytesize > 64.kilobytes
   end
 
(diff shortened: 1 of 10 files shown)


app/models/scenario/user_updates.rb
```diff
@@ -98,6 +98,11 @@ def validate_no_yaml_error
     end
   end
 
+  # Validation method for when a user sets their metadata.
+  def validate_metadata_size
+    errors.add(:metadata, 'can not exceed 64Kb') if metadata.to_s.bytesize > 64_000
```
antw, line 103: Was 64_000 chosen intentionally, rather than 65_536? I doubt this would be an issue in practice, but it might confuse someone that we say the limit is 64Kb but return an error when storing slightly less.

noracato: Good catch!

antw (approved):
Thanks @noracato! I really appreciate that you took care of this. It looks perfect. :+1:

PR #1240, opened 2022-03-29
Set user values for vans when trucks inputs were touched

marliekeverweij:


mabijkerk:
The of the migration plan that deals with technology splits seems to be missing:

> For any truck technology share that is changed by a user e.g. diesel 20% and gasoline 15%:
> - Copy that technology share value to its corresponding share for vans
> - The truck LNG technology share is copied to the van LPG technology share

> For any truck technology share that is not changed by a user:
> - Do nothing

@noracato could you add this to your migration?

noracato:
I'm sorry this isn't finished yet. I ran into the following issue.

Let's say you have your trucks technology split set like this (II3050 - international):
Screenshot 2022-04-04 at 10 12 21

We want to copy all these shares to the vans, looks simple. However, `transport_truck_using_lng_mix_share` was not set to 0 by the user, nor is it in the balanced values, as 0 is it's default value. So if we copy the truck shares one-to-one, instead of zero the start value for `transport_van_using_lpg_mix_share` shows up, and that's 2.4% in this case. So we will have a total of 102.4% the vans.

I tried to solve it in two ways.
1. Using the `Balancer` in ETEngine that is also used for the `ScenarioUpdater`. This makes the migration painfully slow, and somehow does not always produce the right results for me (?).
2. Do a 'quick' balancing. Check if all copied sliders sum to 100, and if so and when there are 'untouched' inputs left in the group, set these remaining inputs to 0 in `balanced_valeus`. This solves the case I used as an example above. But, in ~22% of the cases the copied shares do **not** sum to ~100%. And what do we do then? Do a _nasty_ balance (that is now in the migration) or pull those 22% of scenarios through the Balancer - with the chance of making the migration slow again?

Would you have any ideas @antw? I would appreciate it if you could take a look.

antw:
I can take a longer look later today

> So if we copy the truck shares one-to-one, instead of zero the start value for transport_van_using_lpg_share shows up, and that's 2.4% in this case. So we will have a total of 102.4% the vans.

Is LNG the only offending input? If so, could we instead assume the user has always provided a value of 0.0 for LNG in vans and only set something different if there's a value either in `user_values` or `balanced_values`? Scenarios in which the user has provided no values for trucks could be left alone.

I'm guessing this isn't the whole picture, and perhaps doesn't help in the 22% of difficult scenarios. Can you describe what goes wrong in that 22% of scenarios? Are there several inputs that cause the sum to be something other than 100%?

> I would appreciate it if you could take a look.

I can certainly do so this afternoon.

noracato:
Thanks @antw!

> EDIT: Ok it was a typo I think.

After I fixed the typo it's still happening, sadly.

> Is LNG the only offending input?

Nope. I logged the amount of times the `lng/lpg` input had to be adjusted and it was only in about a fifth of the scenarios. I also considered the idea of always setting it to zero (unless..), but then we have to find out if this is also applicable for other datasets. Maybe Germany does not have 0 as a start value for trucks.

> Can you describe what goes wrong in that 22% of scenarios? Are there several inputs that cause the sum to be something other than 100%?

I can spilt up the 22% in two cases: In 20% of the scenarios there is only one input not copied/untouched, and the sum of the copied inputs is not 100%. And in the remaining two percent of scenarios there are multiple inputs not copied/untouched and the sum of the copied inputs is not 100%.
But I will log some more and go into a litte more detail!

mabijkerk (approved):

PR #1251, opened 2022-05-24
Update the MeritConfigSerializer

noracato:
Adds support for dynamic curves to the `MeritConfigSerializer`, and now also incorporates `Merit::Users`, and most flex options. Except from the interconnectors, the curves that accompany the participants are load profiles. For interconnectors it's their price curve.

Little slow, but does its job.


antw (approved):
Looks good to me. Yeah, it's a little slow but I imagine that's mostly because it has to serialize so many curves?

In my case, it runs in about a second which I feel is perfectly acceptable for such a rarely used endpoint.

PR #1267, opened 2022-08-25
Set various industry CHP inputs using electricity capacity

antw:
Currently, five industry CHPs have their total capacity set by setting the heat output capacity. This will be changed soon in ETSource and ETModel to set the total capacity using the electricity output capacity.

This pull request updates the stored user values to the (normally higher) electricity capacity. These values are dataset-dependent, so the migration will calculate each dataset the first time it encounters it and stores the values to be reused by later scenarios in the migration.

⚠️ **Don't merge this yet!** This PR needs to be paired with a corresponding change in ETSource.

@noracato: If you have the time, a code review would be appreciated.
@marliekeverweij: Will you merge this when you have the ETSource changes ready?

---

Ref quintel/etsource#2716
Closes quintel/etsource#2735
See https://3.basecamp.com/3797409/buckets/24616302/todos/5249810906

### DIFF
--- a/db/migrate/20220825115850_convert_chp_heat_inputs_to_electricity.rb
+++ b/db/migrate/20220825115850_convert_chp_heat_inputs_to_electricity.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'etengine/scenario_migration'
+
+# Updates various industry capacity input values to set electricity output rather than heat output.
+#
+# See https://github.com/quintel/etsource/issues/2716
+class ConvertChpHeatInputsToElectricity < ActiveRecord::Migration[7.0]
+  include ETEngine::ScenarioMigration
+
+  INPUTS = {
+    capacity_of_industry_chp_combined_cycle_gas_power_fuelmix:
+      :industry_chp_combined_cycle_gas_power_fuelmix,
+    capacity_of_industry_chp_engine_gas_power_fuelmix:
+      :industry_chp_engine_gas_power_fuelmix,
+    capacity_of_industry_chp_turbine_gas_power_fuelmix:
+      :industry_chp_turbine_gas_power_fuelmix,
+    capacity_of_industry_chp_ultra_supercritical_coal:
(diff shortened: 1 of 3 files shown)


db/migrate/20220825115850_convert_chp_heat_inputs_to_electricity.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def up
+    migrate_scenarios do |scenario|
+      INPUTS.each do |input, node|
+        begin
+          next if !scenario.user_values.key?(input) || scenario.user_values[input].zero?
+        rescue Psych::DisallowedClass
+          # Ignore older scenarios with illegal values.
+          next
+        end
```
noracato, line 32: The migration failed for me on a copy of the beta database:

```
Caused by:
NoMethodError: undefined method `zero?' for "":String

          next if !scenario.user_values.key?(input) || scenario.user_values[input].zero?
                                                                                  ^^^^^^
```
We could for instance add the `NoMethodError` to the `rescue` block, as this sounds like another old scenario with illegal values.

```suggestion
        begin
          next if !scenario.user_values.key?(input) || scenario.user_values[input].zero?
        rescue Psych::DisallowedClass, NoMethodError
          # Ignore older scenarios with illegal values.
          next
        end
```

noracato (changes requested):
Looks good overall, nicely done with the value cache. Only thing is the migration not completely working for my copy of the beta database for some older scenario(s). But that's a quick fix 🙂

db/migrate/20220825115850_convert_chp_heat_inputs_to_electricity.rb
antw: Thanks for catching this. I tested with a copy of the production DB, but it does indeed appear that one beta scenario has an empty value for an input. I think it's okay to skip the scenario.

I've adjusted the PR to account for this.

noracato (approved):
Looks good to me!

PR #1276, opened 2022-10-04
Fix incorrect FLH for flex technologies

### LINKED ISSUE #1271: Full load hours should not exceed 8760 h/y
When I open the II3050 National governance scenario I obtain the following result for the full load hours as displayed in the `Flexible electricity demand technologies` table:

Screenshot 2022-09-28 at 12 18 23

Screenshot 2022-09-28 at 12 18 13

The full load hours are calculated at 13589.35 h/y. This is odd, since we would expect the full load hours to never exceed 8760 h/y. The underlying [query](https://github.com/quintel/etsource/blob/master/gqueries/output_elements/output_series/table_164_storage_options/energy_heat_flexibility_p2h_heatpump_electricity_flhs.gql) shows:
```ruby
- query = V(energy_heat_flexibility_p2h_heatpump_electricity, full_load_hours)
- unit = hours
```

The question is how we think that the full load hours should be defined and how they are currently defined.
In this case it seems like the full load hours are calculated by dividing the energy output by the **electricity input capacity**:
```ruby
V(energy_heat_flexibility_p2h_heatpump_electricity,"output_of_steam_hot_water/MJ_PER_MWH")/
  V(energy_heat_flexibility_p2h_heatpump_electricity,"heat_output_capacity * electricity_input_conversion * number_of_units")

13,589.346218404788
```

Typically we would expect the energy output to be divided by the energy output capacity, in this case the **heat output capacity**. @antw can you check whether the full load hours are indeed calculated this way in ETEngine? If possible, could you also assess the potential fallout of changing the full load hours calculation for heat technologies.

Together with @marliekeverweij (and possibly others) will can then  assess if and how to update the full load hours method.

antw:
Fixes incorrect FLH assigned to flex technologies when their electricity input conversion is not 1.0.

The problem was that merit calculates the FLH of each participant based on its demand for electricity and input efficiency. We were wrongly dividing that by the input efficiency _again_ when assigning the value back to the node.

I tested with the II3050 National scenario described in #1271 and, fortunately, the only merit order participant affected is the P2H heat pump. All other flex technologies in that scenario were unaffected and had the correct FLH already assigned.

Closes #1271

### DIFF
--- a/app/models/qernel/merit_facade/flex_adapter.rb
+++ b/app/models/qernel/merit_facade/flex_adapter.rb
@@ -103,7 +103,7 @@ module Qernel
 
       # Internal: Sets demand and related attributes on the target API.
       def inject_demand!
-        full_load_hours = participant.full_load_hours / input_efficiency
+        full_load_hours = participant.full_load_hours
 
         full_load_seconds =
           if !full_load_hours || full_load_hours.nan?
@@ -116,7 +116,7 @@ module Qernel
         target_api[:full_load_seconds] = full_load_seconds
 
         target_api.demand =
-          full_load_seconds *
+          (full_load_seconds / input_efficiency) *
           participant.input_capacity_per_unit *
           participant.number_of_units
       end


mabijkerk:
This produces the expected FLH for the P2H heat pump 🙌

However, I don't quite understand what you mean with:

> merit calculates the FLH of each participant based on its demand for electricity and input efficiency

I would expect merit to either calculate the FLH based on the energy input and the input capacity or on the energy output and the output capacity. So for the heat pump, either use the electricity input capacity and the electricity demand, or use the steam hot water capacity and the steam hot water demand. See the example below, where for both methods produce the expected FLH (marked in green) for the P2G and P2H:
Screenshot 2022-10-05 at 17 12 59

I don't quite understand how your method using the input efficiency produces the same result?

antw:
> However, I don't quite understand what you mean with:
> > merit calculates the FLH of each participant based on its demand for electricity and input efficiency

Merit calculates the FLH for flexible technologies. For example, it knows what the electricity input efficiency (conversion) is, and how much electricity is consumed. From that, it can determine the FLH.

When saving this value on the graph we were incorrectly dividing the FLH by the electricity conversion. We _do_ need to divide the amount of electricity consumed in order to calculate the total demand for the node. For example: 1000 electricity consumed / 0.25 = 4000 demand. However, it was wrong to also divide the FLH by the conversion; the FLH of the electricity input and the FLH of the node should be the same.

mabijkerk (approved):

PR #1289, opened 2022-11-01
Bump Merit to quintel/merit@c504a27

### LINKED ISSUE #1284: Blackout hours occur without a demand on the HV shortage node
In [this](https://pro.energytransitionmodel.com/scenarios/961420) scenario there are 3 blackout hours. However, when I go to the `energy_power_hv_network_shortage` node, there is no demand on the node, even though I would expect there to be a demand. I would not expect there to be a blackout hour in any case, since there is roughly 500 GW of dispatchable hydrogen installed.

When I set the solar PV plant with battery system to 0 MW installed capacity, the blackouts disappear, however when I also set the onshore inland wind with battery system to 0 MW, the blackouts reappear. Alternatively, removing all the installed capacity for load shifting seems to remove the blackouts as well. Yet another solution seems to be to install nuclear plants.

I know this is not much to go by, but could you have a look at this please and see if you can find out what's going on here @antw? If necessary I can also spend some more time messing around with the sliders. Let me know what you prefer!

antw:
Fixes that flexible load shifting would sometimes return a negative number when querying how much capacity was available for load shifting. This would result in erroneous blackout hours.

Closes #1284

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,7 +49,7 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
-gem 'quintel_merit', ref: 'ef8106a', github: 'quintel/merit'
+gem 'quintel_merit', ref: 'c504a27', github: 'quintel/merit'
 
 gem 'atlas',         ref: '7ce81f0', github: 'quintel/atlas'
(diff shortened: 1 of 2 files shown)


mabijkerk (approved):
Awesome!

PR #1293, opened 2022-11-08
Update SMR capacity in scenarios

### LINKED ISSUE #1292: We should write a scenario migration for the fix of SMR flh
The full load hours of the SMR have been updated from `7500` to `8322`. For `7500` was no source in ETDataset and the `8322` is based on [this datasheet](https://energy.nl/data/steam-methane-reforming-smr-for-hydrogen-production-2/)

Because of this we would like to write a small scenario migration scaling the slider for installed capacity for SMR: `capacity_of_energy_hydrogen_steam_methane_reformer`

[new value] = [old value] * (7500/8322)

Alexander: _"I think most advanced users want to match H2 supply and demand as well as they can. For SMR, I think not much attention is payed to the input capacity, since it uses gas. If it were an electricity option, I think a lot more attention would be payed to the exact installed capacities and to whether the grid could accommodate it."_

Notifying @AlexanderWirtz & @mabijkerk

antw:
Updates the capacity of steam methane reformers to account for changes in the datasets.

Closes #1292

### DIFF
--- a/db/migrate/20221108095859_update_smr_capacity.rb
+++ b/db/migrate/20221108095859_update_smr_capacity.rb
@@ -1,0 +1,29 @@
+require 'etengine/scenario_migration'
+
+# The full load hours of the SMR have been updated from 7500 to 8322. For 7500 was no source in
+# ETDataset and the 8322 is based on this datasheet
+#
+# This migration scales SMRs in existing scenarios to account for the change.
+#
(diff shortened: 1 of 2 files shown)


mabijkerk (approved):
LGTM!

marliekeverweij (approved):

PR #1295, opened 2022-11-10
Fix price set to VoLL when the most expensive dispatchable is used to meet flexible demand

### LINKED ISSUE #1286: Electricity price is set at VoLL unexpectedly
In [this](https://pro.energytransitionmodel.com/scenarios/962204) scenario some very strange price-setting behaviour occurs. It is a scenario that uses quite a lot of our more advanced features, like DSR in industry, external coupling and availability curves for interconnectors.

The strange behaviour is that in unexpected hours the electricity price is set at the VoLL of 3000 euros/MWh, which suggests that the total supply in those hours is insufficient to cover the total demand. However, when I look at those hours, this does not seem to be the case.

**Example for hour 15**
The prices on the interconnectors are the following:
image

The interconnector to Norway is importing, all the others are exporting. The only flexible power plants are the waste CHPs (with and without CCS) and the hydrogen power plants (turbine and CCGT). The hydrogen turbine is not running at full capacity and has a lower price than the other interconnectors (DE, BE, DK and UK), which are priced at 150 euros/MWh. I would therefore expect the hydrogen turbine to set the price at 135.13 euro/MWh.

Would you be able to have a look at this please @antw? Perhaps it has something to do with the effect of availability curves on the calculation of the intersection of total demand and supply?

antw:
Merit has been adjusted so that the price is only set to €3000 when there is insufficient capacity to meet _inflexible_ demand. Previously, if the most expensive dispatchable/flexible producer was fully-loaded, we'd treat that as a deficit. That's not correct as we now have flexible consumers; it's possible for the most expensive dispatchable to be fully loaded without this being a loss of load event.

These screenshots were taken from a local scenario, and may not exactly match the version on production. In my scenario, the number of hours at €3000 exactly matches the number of blackout hours (which is to be expected since a blackout is inflexible demand > production).

| Before | After |
| :---: | :---: |
| b-1 | a-1 |
| b-2 | a-2 |
| b-3 | a-3 |
| b-jan | a-jan |

Closes #1286

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,7 +49,7 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
-gem 'quintel_merit', ref: 'c504a27', github: 'quintel/merit'
+gem 'quintel_merit', ref: '280cb72', github: 'quintel/merit'
 
 gem 'atlas',         ref: '7ce81f0', github: 'quintel/atlas'
 gem 'fever',         ref: 'bf092b2', github: 'quintel/fever'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -26,8 +26,8 @@ GIT
 
 GIT
   remote: https://github.com/quintel/merit.git
-  revision: c504a2704741837e8ce1ce1baa1b1b71114e0890
-  ref: c504a27
+  revision: 280cb72ce8d427746f36e78832366ba9fd2bb5a3
+  ref: 280cb72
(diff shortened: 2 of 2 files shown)


mabijkerk:
Very nice work! When I run the scenario locally I get don't get any blackout hours, nor any hours where the price is 3000 euros/MWh.

mabijkerk (approved):

PR #1311, opened 2023-01-12
Add the ability to delete your account

antw:
Account deletion happens in the background (since a user may have many scenarios, access tokens, etc) and needs to also tell ETModel to delete its data about the user.

Deleted users are flagged with a `deleted_at` column, which prevents them from being able to sign in through the web interface. The user will be completely removed by the DeleteUserJob background job.


mabijkerk (approved):

antw:
Thank you once again for the translations, @mabijkerk. Very much appreciated!

PR #1317, opened 2023-01-27
Remove aggregate inputs

mabijkerk:
This migration concerns the following keys:

- [industry_useful_demand_for_chemical_aggregated_industry](https://github.com/quintel/etsource/blob/master/inputs/demand/industry/industry_useful_demand_for_chemical_aggregated_industry.ad)
- [industry_useful_demand_for_chemical_aggregated_industry_electricity_efficiency](https://github.com/quintel/etsource/blob/master/inputs/demand/industry/industry_useful_demand_for_chemical_aggregated_industry_electricity_efficiency.ad)
- [industry_useful_demand_for_chemical_aggregated_industry_useable_heat_efficiency](https://github.com/quintel/etsource/blob/master/inputs/demand/industry/industry_useful_demand_for_chemical_aggregated_industry_useable_heat_efficiency.ad)

These keys stem from a time when the fertilizer sector was not explicitly modelled for every dataset. This was the case in datasets where `has_aggregated_chemical_industry = true`.

When these aggregate inputs are set in combination with the individual inputs of the fertilizer and other chemical sector however, the demand from the fertilizer sector goes to 0. This is not intentional and can present problems for scenarios where these inputs have been set. 

Since all datasets currently have an explicit fertilizer sector and the aggregate inputs are not accessible in the frontend, we want to remove the aggregate inputs from all recent scenarios (>2021). We can then remove the inputs from ETSource altogether.

@noracato when I check the presence of each slider in migratable scenarios I get 682, 1564 and 1564 cases. However, when I run the migration I get 7663 migrated scenarios. Could you look into this to see where this difference comes from?

Merging is on hold until cleared by ENTSO.

### DIFF
--- a/db/migrate/20230125153925_remove_redundant_aggregate_inputs.rb
+++ b/db/migrate/20230125153925_remove_redundant_aggregate_inputs.rb
@@ -11,9 +11,9 @@ class RemoveRedundantAggregateInputs < ActiveRecord::Migration[7.0]
 
 
   def up
-    migrate_scenarios(since: Date.new(2021, 1, 1)) do |scenario|
+    migrate_scenarios(since: Date.new(2022, 1, 1)) do |scenario|
       REDUNDANT_AGGREGATE_KEYS.each do |key|
-        scenario.user_values.delete(key)
+        scenario.user_values.delete(key) if scenario.user_values.include? key
       end
     end
   end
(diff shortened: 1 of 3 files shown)


mabijkerk:
The inputs discussed in this PR have long been removed from ETSource. This PR is no longer relevant.

PR #1323, opened 2023-03-03
Updates to the costs data-export

mabijkerk:
There are a number of issues with the costs data-export, as described in https://github.com/quintel/etsource/issues/2859:

Goes with https://github.com/quintel/etsource/pull/2860

Fixes include the following:

- Structure of totals is changed slightly so that Group totals are clearly defined
- Hydrogen storage costs is now assigned the right query, so that a value is displayed
- CO2 costs are now assigned the right queries, so that values are displayed
- Heat infrastructure costs query is now assigned the right query, so that double-counting with the heat storage costs is avoided
- Hydrogen infrastructure costs total query is now removed from the data-export, so that double-counting with the two underlying queries that are also in the data-export is avoided

Fixes do **not** include the following:

- All CCS costs are included in the CO2 costs queries in the row "Total ex ccs". Conceptually this does not make sense and it should be addressed

### DIFF
--- a/app/serializers/costs_parameters_serializer.rb
+++ b/app/serializers/costs_parameters_serializer.rb
@@ -147,7 +147,8 @@ class CostsParametersSerializer
   # If no subgroup was supplied, gets the query with the name of the main group
   def group_total_row(group, subgroup = nil)
     query = subgroup ? "#{group}_#{subgroup}" : group.to_s
-    query_row(group, subgroup, query, 'Total')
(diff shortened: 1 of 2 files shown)


mabijkerk:
To do for new issue:

- [ ] Review generic structure (totals are displayed in a way that is not very practical, some subgroups only have totals)
- [ ] Change column that is called 'ex CCS' even though CCS is included in the final group
- [ ] Provide limited documentation for interpretation of the data-export

DorinevanderVlies (approved):

PR #1335, opened 2023-04-11
Allow updating ESDL file attachment on a scenario

### LINKED ISSUE #1334: ESDL file attachments don't allow updating
Putting an ESDL file twice on the same scenario returns the following error message:

image

We should investigate why this started happening.

References quintel/etm-esdl#94

noracato:
I found we were explicitly not allowing updating the file. @redekok do you remember why we chose to do that? 

This PR allows for the file to be updated through the ESDL files API endpoint.

Closes #1334 
References quintel/etm-esdl#94

### DIFF
--- a/app/models/file_upload_handler.rb
+++ b/app/models/file_upload_handler.rb
@@ -18,7 +18,7 @@ class FileUploadHandler
 
   # Public: attaches the file to the scenario as content type xml
   def call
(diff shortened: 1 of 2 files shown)


redekok:
> @redekok do you remember why we chose to do that?

No clue.. 😅 

But thanks for fixing this so quickly!

PR #1342, opened 2023-05-15
Coupling transparancy: show coupling of scenario, inputs and uncouple

### LINKED ISSUE #1340: Pass `coupled` as a scenario parameter in the API
For the coupling project, certain inputs will be marked as part of a coupling. When one of these inputs is set a scenario should be marked as `coupled`. This attribute needs to be available to pass through the API towards ETModel.

For now we don't care which coupling is active (as we only have the one from the CTM) but let's keep this option open for the future when developing this!

- [ ] On `Scenario` add a new method `coupled`
- [ ] Incorporate this attribute in the serializer

### LINKED ISSUE #1341: In inputs serializer, show if disabled input is part of coupling
When serializing the inputs for the inputs endpoint, don't only show if the input is `disabled` or who it's `disabled_by`, but also show if that was part of a coupling. 

- [ ] Check if the slider that is has been `disabled_by` is part of a coupling group
- [ ] Add Boolean attribute `part_of_coupling`

noracato:
Adds support for a more transparant coupling witrh other models:
- A new attribute for scenarios the is available through the scenario API (closes #1340)
- Show if inputs that can be disabled are part of a `coupling_group` though the inputs API (closes #1341 )
- Uncouple a scenario by passing `coupling: false` in the scenario endpoint

I made it so that there is space in the future to communicate which coupling the scenario is part of. Instead of a `true`/`false` we could communicate the actual coupling groups.

---
Goes with
quintel/atlas#158
quintel/etmodel#4114
quintel/etsource#2902

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -77,7 +77,7 @@ gem 'ruby-progressbar'
 # own gems
 gem 'quintel_merit', ref: 'c722536', github: 'quintel/merit'
 
-gem 'atlas',         ref: '0dad20e', github: 'quintel/atlas'
+gem 'atlas',         ref: '6ea3589', github: 'quintel/atlas'
 gem 'fever',         ref: 'bf092b2', github: 'quintel/fever'
 gem 'refinery',      ref: '72eacf8', github: 'quintel/refinery'
 gem 'rubel',         ref: 'e36554a', github: 'quintel/rubel'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,8 +6,8 @@ GIT
 
 GIT
   remote: https://github.com/quintel/atlas.git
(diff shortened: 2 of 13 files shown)


mabijkerk:
I tested the functionality in the frontend and using the API. It seems to work well, great work! ✨ 

One question that I do have is what happens if you send a `coupling: true` request to the scenario using the API?

noracato:
Thanks for your review!

> A question that I do have is what happens if you send a coupling: true request to the scenario using the API?

Nothing, it's completely ignored. Do we have to specify that in the docs?

> And one potential improvement that I spotted is that the API request coupling: false does not seem to hide the coupling button in the frontend

I'll check that in ETModel! ETEngine is returning it correctly.

mabijkerk (approved):

PR #1343, opened 2023-05-22
Fix storage optimisation algorithm

### LINKED ISSUE #1300: Flow battery output capacity seems to be capped below specified output capacity
**Issue**
In [this](https://energytransitionmodel.com/scenarios/951061) scenario, there is 4500 MW of input capacity installed for `energy_flexibility_flow_batteries_electricity`. Given that the electricity output capacity as specified by the [node](https://github.com/quintel/etsource/blob/master/graphs/energy/nodes/energy/energy_flexibility_flow_batteries_electricity.converter.ad) is 50% lower than the typical input capacity, the output capacity is 2250 MW:
``` ruby
V(energy_flexibility_flow_batteries_electricity,"number_of_units * electricity_output_capacity")

2,250.0
```

Throughout the year, there are hours where the maximum input capacity is used:
```ruby
MAX(V(energy_flexibility_flow_batteries_electricity,electricity_input_curve))

4,500.0000000000655
```

However, when I look at the output curve of the battery, the maximum output capacity is never used:
```ruby
MAX(V(energy_flexibility_flow_batteries_electricity,electricity_output_curve))

1,102.5000000000355
```

The strange thing is that the 1102 MW isn't a sporadic event, based on the output curve it seems like a hard cap on the output capacity. Below the charging behaviours is visualised for the first few hours in the year:

Screenshot 2022-12-09 at 11 42 48

**Question**
@antw would you able to look into this issue and find out what is happening here? I thought it might be related to the distribution of electricity between multiple storage technologies with the forecasting algorithm on, but if I remove the installed capacity of the other storage technologies, the maximum output capacity of the flow batteries does not change.

noracato:
New implementation of output efficiency, so it is not applied twice after calculating the curve.

Finally closes #1300

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -75,7 +75,7 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
-gem 'quintel_merit', ref: 'e8a9494', github: 'quintel/merit'
+gem 'quintel_merit', ref: '3b60eed', github: 'quintel/merit'
 gem 'atlas',         ref: '9b616c4', github: 'quintel/atlas'
 gem 'fever',         ref: 'bf092b2', github: 'quintel/fever'
 gem 'refinery',      ref: '72eacf8', github: 'quintel/refinery'
--- a/Gemfile
+++ b/Gemfile
@@ -75,7 +75,7 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
-gem 'quintel_merit', ref: '67ab37b', github: 'quintel/merit'
+gem 'quintel_merit', ref: 'c16d03b', github: 'quintel/merit'
 gem 'atlas',         ref: '9b616c4', github: 'quintel/atlas'
 gem 'fever',         ref: 'bf092b2', github: 'quintel/fever'
 gem 'refinery',      ref: '72eacf8', github: 'quintel/refinery'
--- a/Gemfile
+++ b/Gemfile
@@ -75,7 +75,7 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
-gem 'quintel_merit', ref: '3b60eed', github: 'quintel/merit'
+gem 'quintel_merit', ref: '67ab37b', github: 'quintel/merit'
 gem 'atlas',         ref: '9b616c4', github: 'quintel/atlas'
 gem 'fever',         ref: 'bf092b2', github: 'quintel/fever'
 gem 'refinery',      ref: '72eacf8', github: 'quintel/refinery'
--- a/Gemfile
+++ b/Gemfile
@@ -75,7 +75,7 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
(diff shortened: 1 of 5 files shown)


thomas-qah:
Hey @noracato! I checked the `battery-opt` branch out and ran `rspec` locally. Doing this also returned 5 errors:

```
3205 examples, 5 failures, 20 pending

Failed examples:

rspec ./spec/models/qernel/merit_facade/storage_optimization_spec.rb:216 # Qernel::MeritFacade::StorageOptimization#reserve_for with a single battery with an output efficiency of 0.8 calculates the battery reserve
rspec ./spec/models/qernel/merit_facade/storage_optimization_spec.rb:223 # Qernel::MeritFacade::StorageOptimization#reserve_for with a single battery with an output efficiency of 0.8 calculates the battery load
rspec ./spec/models/qernel/merit_facade/storage_optimization_spec.rb:238 # Qernel::MeritFacade::StorageOptimization#reserve_for with a single battery with an output efficiency of 1.2 calculates the battery reserve
rspec ./spec/models/qernel/merit_facade/storage_optimization_spec.rb:245 # Qernel::MeritFacade::StorageOptimization#reserve_for with a single battery with an output efficiency of 1.2 calculates the battery load
rspec ./spec/requests/api/v3/saved_scenarios_spec.rb:107 # Saved scenarios API when fetching a single saved scenario when the saved scenario exists but the scenario does not has no data for the scenario
```

Let's discuss this thursday/friday :)

mabijkerk:
Work out expected results based on test numbers in error message of @thomas-qah

mabijkerk:
I have not yet managed to fully reproduce and understand the tests. However, I did already notice an unexpected result for the first test in rb:218:

```ruby
       expected: [5000.0, 4250.0, 3500.0, 2750.0, 2000.0, 750.0, 1500.0, 2250.0, 3000.0, 3750.0, 4000.0, 5000.0]
            got: [5000.0, 4500.0, 3750.0, 3000.0, 2250.0, 1000.0, 1750.0, 2500.0, 3000.0, 3750.0, 4000.0, 5000.0]
```

The drop in reserve from 2250.0 to 1000.0 should not be possible, because this indicates a discharge of 1250.0 and the specs of the battery for this test state that the capacity should be 1000.0:

```ruby
    context 'with a single battery with an output efficiency of 0.8' do
      let(:adapters) do
        [
          consumer_double(:must_run, (([15_000] * 5 + [30_000]) + ([5_000] * 6 )) * 365),
          battery_double(key: :a_battery, volume: 5000.0, capacity: 1000.0, efficiency: 0.8)
        ]
      end
```

mabijkerk:
The difference between running the spec locally on Mac with respect locally on Linux or on Semaphore was caused by an operating system difference in the `sort` method of Ruby. Non-unique values are sorted in a different way by both operating system. To counter this we have implemented a second-level attribute to the `sort` function, that sorts on the index as well, and therefore guarantees a stable result.

mabijkerk (approved):
🔋

PR #1358, opened 2023-09-15
Handle inputs with boolean values correctly

### LINKED ISSUE #1321: Discrete inputs handled as continious inputs
There are several discrete values in the scenario inputs that have a min and max value, which probably should have a permitted_values list similar to the 'settings_weather_curve_set' input. It is now possible to assign these boolean like inputs with strange values like 0.5 that in some cases are scenario breaking.

            'heat_storage_enabled',
            'merit_order_subtype_of_energy_power_nuclear_uranium_oxide',
            'settings_enable_merit_order',
            'settings_enable_storage_optimisation_energy_flexibility_hv_opac_electricity',
            'settings_enable_storage_optimisation_energy_flexibility_pumped_storage_electricity',
            'settings_enable_storage_optimisation_energy_flexibility_mv_batteries_electricity',
            'settings_enable_storage_optimisation_energy_flexibility_flow_batteries_electricity',
            'settings_enable_storage_optimisation_transport_car_flexibility_p2p_electricity',

Being able to recognize these boolean-like inputs directly would make it easier to interpolate the scenario and get some additional feedback from the engine when passing an invalid number.

thomas-qah:
## What?
This PR adds validation for inputs that contain boolean values. It also makes sure the values for those inputs are skipped values for transition paths are interpolated.

## Why?
As described in [this](https://github.com/quintel/multi-year-charts/issues/31) and [this](https://github.com/quintel/etengine/issues/1321) issue boolean values are currently not handled properly: they were not validated properly and their value was interpolated and recalculated for scenarios in transition paths.

## How?
- Validation for inputs with boolean values is now added to the `Api::V3::ScenarioUpdater`
- Interpolation for inputs with boolean values is now skipped in the `Scenario::YearInterpolator`

Closes #1321
Closes quintel/multi-year-charts#31

### DIFF
--- a/spec/models/scenario/year_interpolator_spec.rb
+++ b/spec/models/scenario/year_interpolator_spec.rb
@@ -94,6 +94,35 @@ RSpec.describe Scenario::YearInterpolator do
     end
   end
 
+  context 'with a scenario containing a boolean input' do
+    let(:source) do
+      FactoryBot.create(:scenario, {
+        id: 99999,
+        end_year: 2050,
+        user_values: {
+          'grouped_input_one' => 75,
+          'bool' => 0
+        }
+      })
+    end
+
+    let(:interpolated) { described_class.call(source, 2030) }
+
+    it 'sets the inputs' do
+      expect(interpolated.user_values.keys.sort)
+        .to eq(%w[bool grouped_input_one])
+    end
+
+    it 'interpolates the numeric inputs' do
+      expect(interpolated.user_values['grouped_input_one'])
+        .to be_within(1e-2).of(87.82)
+    end
+
+    it 'retains the value of the boolean input' do
+      expect(interpolated.user_values['bool']).to eq(0)
+    end
+  end
+
   context 'with a year older than the analysis year' do
     let(:source) do
       FactoryBot.create(:scenario, {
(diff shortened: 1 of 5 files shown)


spec/models/scenario/year_interpolator_spec.rb
```diff
@@ -94,6 +94,35 @@
...
   end
 
+  context 'with a scenario containing a boolean input' do
+    let(:source) do
+      FactoryBot.create(:scenario, {
+        id: 99999,
+        end_year: 2050,
+        user_values: {
+          'grouped_input_one' => 75,
+          'bool' => 0
```
noracato, line 104: This test would be more useful when we test for a future value of 1, given that the present value is always 0. Then we can see whether the interpolation will keep the value at 1 for the intermediate scenario.

noracato (commented):
Nice and quickly fixed! Thanks @thomas-qah 

I left one comment on the test for the interpolator. Could you maybe also add a test for the new `bool` input [here](https://github.com/quintel/etengine/blob/339cf7db2770b972c92653f4e7fbc36a9c1d0315/spec/requests/api/v3/update_input_spec.rb#L701)?

noracato (approved):
Looks great! 🎉

PR #1379, opened 2023-12-14
Update costs data-export for heat

mabijkerk:
Updates the costs data-export to include temperature levels for heat storage.


KoenvanB (changes requested):
Approved, except the comment in line 13 of costs_parameter_serializer_spec needs to be updated to "14 queries"
(although we are probably going to revise this again later anyway)

PR #1380, opened 2023-12-15
Introduce more fine-grained load-shifting

thomas-qah:
## What?
This PR adds support for setting up more fine-grained load-shifting by our users.

## Why?
The currently existing load-shifting functionality caused too much of a peak in the demand for the electricity system. The user needed a way to spread out this demand more evenly over time.

## How?
By adding support to set and influence `output_capacity` in etengine.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -77,7 +77,7 @@ gem 'ruby-progressbar'
 
 # own gems
 gem 'quintel_merit', ref: '421f3fb', github: 'quintel/merit'
-gem 'atlas',         ref: '5e85101', github: 'quintel/atlas'
+gem 'atlas',         ref: '2db914b', github: 'quintel/atlas'
 gem 'fever',         ref: 'bf092b2', github: 'quintel/fever'
 gem 'refinery',      ref: '72eacf8', github: 'quintel/refinery'
 gem 'rubel',         ref: 'e36554a', github: 'quintel/rubel'
--- a/Gemfile
+++ b/Gemfile
@@ -77,8 +77,7 @@ gem 'ruby-progressbar'
 
 # own gems
 gem 'quintel_merit', ref: '421f3fb', github: 'quintel/merit'
-#gem 'atlas',         ref: 'f7aebb9', github: 'quintel/atlas'
-gem 'atlas',         path: '../atlas'
+gem 'atlas',         ref: 'f7aebb9', github: 'quintel/atlas'
 gem 'fever',         ref: 'bf092b2', github: 'quintel/fever'
 gem 'refinery',      ref: '72eacf8', github: 'quintel/refinery'
 gem 'rubel',         ref: 'e36554a', github: 'quintel/rubel'
--- a/Gemfile
+++ b/Gemfile
@@ -77,7 +77,8 @@ gem 'ruby-progressbar'
 
 # own gems
 gem 'quintel_merit', ref: '421f3fb', github: 'quintel/merit'
-gem 'atlas',         ref: 'f7aebb9', github: 'quintel/atlas'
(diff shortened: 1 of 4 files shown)


app/models/qernel/node.rb
```diff
@@ -415,25 +415,25 @@ def calculate
   # @return [Float] The demand of this node
   #
   def update_demand
```
noracato, line 417: Hi! This is very dangerous to do. Like this we will not know if the modelling was wrong or insufficient, the engine will just silently break and energy flows might not add up without us knowing. Please revert all changes in this file!

Gemfile
```diff
@@ -77,7 +77,7 @@ gem 'ruby-progressbar'
 
 # own gems
 gem 'quintel_merit', ref: '421f3fb', github: 'quintel/merit'
-gem 'atlas',         ref: 'f7aebb9', github: 'quintel/atlas'
+gem 'atlas',         ref: '5e85101', github: 'quintel/atlas'
```
noracato, line 80: FYI: For updating Atlas we always use the same commit message: "Bump Atlas to quintel/atlas@5e85101"

Also if you work with Atlas locally, we normaly push Atlas and update the ref here instead of pushing your path 🙂

noracato (changes requested):

app/models/qernel/merit_facade/load_shifting_adapter.rb
```diff
@@ -79,12 +79,12 @@ def limiting_curve
...
 
       def input_capacity
-        # The input capacity will be further affected in Merit by the availability of the node.
-        @input_capacity ||= limiting_curve.max
+        @input_capacity = (@config.input_capacity_from_share || 0.0) * output_capacity
       end
 
       def output_capacity
-        input_capacity
+        # The input capacity will be further affected in Merit by the availability of the node.
```
noracato, line 86: Is this comment still applicable here?

Either this should now be a comment about the output capacity, or it should be moved to the `input_capacity` method. Check Merit to see which one is the correct option.

app/models/qernel/node.rb
noracato, line 1: Let's try to leave all changes to this file out of this PR, as it has nothing to do with the added functionality!

noracato (changes requested):
Since there are only a few lines of code that are changed in this PR, maybe it's nice to remake it into just one commit, instead of the 6 we will have otherwise 🙂

app/models/qernel/merit_facade/load_shifting_adapter.rb
thomas-qah: I checked the (load_shifting module)[https://github.com/quintel/merit/blob/master/lib/merit/flex/load_shifting.rb] in Merit and to me it seems like neither the input nor the output capacity is changed or calculated there.  It does seem like those are used to calculated the (future) availability however.

Should I remove the comment altogether?

app/models/qernel/node.rb
thomas-qah: They seemed like useful changes to me either way, but I reverted them.

app/models/qernel/merit_facade/load_shifting_adapter.rb
mabijkerk: I think the confusion was mainly about the comment referring to the input capacity in the definition of the output capacity. I have updated the comments.

mabijkerk (approved):
⚡

thomas-qah:
> Since there are only a few lines of code that are changed in this PR, maybe it's nice to remake it into just one commit, instead of the 6 we will have otherwise 🙂

@noracato: If we 'Squash and merge' this branch it will end up as one commit in `master`, which is exactly what you want, right?

PR #1402, opened 2024-02-07
Update costs parameters serializer

redekok:
- Restructure the format of the cost export based on the FiT project design
  - Refactor `CSV_HEADERS`
  - Update methods for node rows
  - Update queries for query rows
- Make sure the cost export is complete in terms of relevant node and query rows
  - Update static `QUERIES` variable to match the project design
  - Add liquid hydrogen and lohc to the `costs_carriers` group
  - Add non-existing subgroups from the `costs_co2` group and add ccus 
- Clean up some of the old code 
  - Remove subtotals per (sub)group
  - Remove transport from the `costs_building_and_installations` group


app/models/qernel/node_api/cost.rb
```diff
@@ -222,6 +222,17 @@ def total_costs
...
       end
 
+      # Internal: Calculates the total cost excluding fuel costs of a plant in euro per plant per year.
+      #
+      # Total cost is made up of OPEX and CAPEX (excluding fuel costs).
+      #
+      # Returns the total costs excluding fuel costs for one plant per year.
+      def total_costs_excluding_fuel_costs
+        fetch(:total_costs_excluding_fuel_costs) do
+          capital_expenditures + operating_expenses if capital_expenditures && operating_expenses
```
mabijkerk, line 232: I don't see `capital_expenditures` and `operating_expenses` defined anywhere in ETEngine. Furthermore, `total_costs` are built using `fixed_costs + variable_costs`, while `total_costs_excluding_fuel_costs` uses `capital_expenditures + operating_expenses`. I would then only expect the variable_costs component to be changed.

app/serializers/costs_parameters_serializer.rb
```diff
@@ -5,46 +5,135 @@
...
+    ],
+    costs_carriers_heat: %w[
+      costs_carriers_heat
+    ],
+    costs_carriers_natural_gas: %w[
+      costs_carriers_natural_gas
+    ],
+    costs_carriers_waste: %w[
+      costs_carriers_waste
     ],
```
mabijkerk, line 69: Is there a reason why all these carriers have a separate group, instead of putting under the same generic group `costs_carriers`.

app/serializers/costs_parameters_serializer.rb
```diff
@@ -5,46 +5,135 @@
...
   }.freeze
 
+  CSV_HEADER = [
+    'Group', 'Subgroup', 'Key',
+    'Total costs (eur)', 'Total CAPEX (eur)', 'Total OPEX (eur)',
+    'Capital costs (eur)', 'Depreciation costs (eur)',
+    'Fixed operation and maintenance costs (eur)', 'Variable operation and maintenance costs (eur)',
+    'Total investment over lifetime (eur)', 'WACC', 'Construction time (years)', 'Technical lifetime (years)',
+    'CO2 emission costs (eur)', 'Number of units'
+  ]
```
mabijkerk, line 127: If there are costs that specify "Total investment over lifetime (eur)", shouldn't all other costs then be specified as "(eur/y)"?

app/serializers/costs_parameters_serializer.rb
```diff
@@ -244,7 +356,7 @@ def costs_storage_and_conversion
   # This group only needs subtotal group queries
```
mabijkerk, line 356: What I don't understand is why still a difference is made between `groups_with_subtotal` and `groups_without_subtotal`. It was my understanding that all subtotals were removed?

app/models/qernel/node_api/cost.rb
KoenvanB: @redekok Could it be that this was an aborted attempt to write a method, but that we ultimately opted for a manual, hard coded calculation in lines 213-214?

PR #1409, opened 2024-03-19
BECCUS

kaskranenburgQ:
With this update, the option to model BECC(U)S plants with the ETM becomes possible.

For the ETengine repository this means the following changes have been implemented:
- Addition of biogenic co2 capture cost method to cost calculations.
- Addition of biogenic co2 capture costs to the cost export.


mabijkerk (changes requested):
To the existing dispatchable biomass CHP, a must-run biomass CHP is added. This means that the following inputs need to be migrated:

- capacity_of_energy_chp_local_wood_pellets to capacity_of_energy_chp_local_wood_pellets_dispatchable
- share_of_energy_chp_local_ht_wood_pellets to share_of_energy_chp_local_ht_wood_pellets_dispatchable
- share_of_energy_chp_local_mt_wood_pellets to share_of_energy_chp_local_mt_wood_pellets_dispatchable

@kaskranenburgQ do agree with this proposed migration? Am I missing any other sliders that need to be migrated?

mabijkerk:
I added a migration that takes renames the relevant inputs. @noracato could you check this migration?

app/serializers/costs_parameters_serializer.rb
```diff
@@ -211,7 +211,8 @@ def node_rows_for_subgroup(graph, group, subgroup)
         subgroup, # Subgroup
         node.key, # Key
         format_num(node.query.total_costs_per(:node) - 
-          node.query.fuel_costs_per(:node) - node.query.co2_emissions_costs_per(:node)), # Total costs (eur)
+          node.query.fuel_costs_per(:node) - node.query.co2_emissions_costs_per(:node) - 
+          node.query.captured_biogenic_co2_costs_per(:node)), # Total costs (eur)
```
mabijkerk, line 215: This method still produces an error when downloading the costs export. It seems that the AREA attribute `captured_biogenic_co2_price` is not found in the molecule graph, which is strange, because the AREA attribute `co2_price` works generally the same way.

mabijkerk (approved):

PR #1415, opened 2024-04-03
Gquery API

noracato:
Adds an API endpoint to collect all gqueries.

```
GET api/v3/gqueries
```

Accepts `labels` parameter containing a list of labels to filter on. The filter work cumulative, so passing `heat` and `final_demand` retrieves all gqueries with either a label final_demand or heat.


mabijkerk:
@noracato for accessibility I added the GQuery navigator to the subnav bar like this:
Screenshot 2024-04-05 at 13 23 48

mabijkerk (changes requested):
When I try to open a scenario on etmodel, the scenario won't load and I receive the following error in the Developer console:

Screenshot 2024-04-05 at 13 55 09

mabijkerk:
When I run the script below in my terminal, I expect only a very limited number of gqueries to be returned. Instead, I get the full list. Is my script faulty or is the filter not working properly? FYI - I do include the local url and my personal token.

```python
labels = {
  "labels":	
    [  
      "bezier_178_source_of_heat_in_industry_chemical_other"
    ]
}

response = requests.get(ete_url, headers=headers, params=labels)

if response.status_code == 200:
    print(response.text)
else:
    print(f"Request failed with status code {response.status_code}")
```

noracato:
Thanks for the review!

> When I try to open a scenario on etmodel,

That was because in the mean time multi user was merged. A simple rebase fixed this.

>  Is my script faulty or is the filter not working properly?

It's a JSON API, so instead of `params` you send `json` parameters:

```
response = requests.get(ete_url, json=labels)
```

I tested you script with this change and I do get the correct results!

So actually I think we can merge, if you are OK?

mabijkerk (approved):
I tested it using your suggestion and it works! 🚀

mabijkerk:
@noracato do we need to update the API docs as well? Should I ask someone to write it?

Issue #178, opened 2011-11-07
Make flexible links also work without demand


WvanLelyveld:
Currently, the flexible links are only correctly calculated when they actually distribute a demand.

Without a demand, it is possible that the share and flexible link will not add up to 1, therefore creating an incorrect energy balance. This is not correct and could possible produce errors when these link shares are queried through the GQL.

This is how is looks without a demand:
![without demand](http://f.cl.ly/items/1s41372d1K1k36350j2K/Image%202011.11.07%202:57:40%20PM.png) As you can see the input shares do not add up to 1.

With a demand it is correct:
![with demand](http://f.cl.ly/items/01221A1r053O3i1r2p3q/Image%202011.11.07%202:56:52%20PM.png)

PR #1418, opened 2024-04-08
GQL

kaskranenburgQ:
Improves GQL function INVALID_TO_ZERO

Goes together with PR [Add GQL documentation #178](https://github.com/quintel/documentation/pull/178#issuecomment-2042040247)

### DIFF
--- a/app/models/gql/runtime/functions/legacy.rb
+++ b/app/models/gql/runtime/functions/legacy.rb
@@ -39,14 +39,14 @@ module Gql::Runtime
         first.flatten - last.flatten
       end
 
-      def INVALID_TO_ZERO(*keys)
+      def INVALID_TO_ZERO(*values)
         is_invalid = Proc.new {|v| v.nil? || (v.respond_to?(:nan?) && v.nan?) }
 
         if values.respond_to?(:map)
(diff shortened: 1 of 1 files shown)


mabijkerk:
I have not seen this GQL function being used in any gqueries and I was not familiar with it myself. Unless if we can come up with a relevant use case, I would conclude that we should remove this GQL instead. Do you agree @kaskranenburgQ? Can we simply remove a GQL function @noracato?

kaskranenburgQ:
I think this function could be useful if you are preparing data.
I don't think it is therefore necessary in GQL, since with our use of the language no data needs to be prepared. 
Removing instead of fixing the function would be a better idea in that case.

noracato:
> Can we simply remove a GQL function @noracato?

If it's not used in any queries, you should be safe to remove it. Make sure to check the `specs` to see if you should also remove the tests for it!

> Removing instead of fixing the function would be a better idea in that case.

Agreed! Always nice when we can declutter!

mabijkerk:
@kaskranenburgQ can you remove the GQL function and any accompanying specs from ETEngine and the documenation?

PR #1424, opened 2024-04-30
Tag scenario versions

noracato:
Allows for scenarios to be tagged as a version. The version keeps track of the person who last changed the scenario, and allows for a description for the specific version. Each scenario is only allowed one tag.

This PR goes together with quintel/etmodel#4251.

Still todo:
- [ ] when scenario is updated, check if it was tagged and update the user with the current user


kndehaan:
I think this works as expected.

kndehaan (approved):
To my opinion, it works as desired.

PR #1427, opened 2024-05-07
GQL function percentage change per year

kndehaan:
This new function is introduced for solving the issue as specified in this [PR](https://github.com/quintel/etsource/pull/2970). 

Related to the function UPDATE() with update_type = %y (relative_per_year). Now, the same result can be obtained using a GQL function rather than specifying the update_type of an input. 

Goes with:
- https://github.com/quintel/etsource/pull/2970
- https://github.com/quintel/etengine/pull/1427

Next step (when PR is approved):
- [ ] Include new functions in GQL documentation

### DIFF
--- a/app/models/gql/runtime/functions/legacy.rb
+++ b/app/models/gql/runtime/functions/legacy.rb
@@ -137,6 +137,13 @@ module Gql::Runtime
         values.first
       end
 
+      # Calculates the result of first value multiplied with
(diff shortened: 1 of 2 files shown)


kndehaan:
More effort is required to solve this issue, therefore this PR will be closed and proposed next steps are described in this issue https://github.com/quintel/etsource/issues/2967

PR #1439, opened 2024-06-10
Added a ctrl + enter shortcut for editing scenarios

### LINKED ISSUE #1436: Add shortcut command for updating scenario in engine
When going to the GQL Sandbox in the engine, there is a very useful feature that runs a GQL with CTRL+RETURN.

Screenshot 2024-06-05 at 09 35 27

When editing a current scenario, it would be very useful if the "Update Scenario" button has the same shortcut command. Now I need to scroll all the way down before being able to click the button.

Screenshot 2024-06-05 at 09 36 25

This is something that @louispt1 may pick up at some point.

louispt1:
Generalised the CodeMirror function in application.js so any codemirror instances could use the ctrl+enter command to submit a form. Hence editing scenarios can be submitted with ctrl + enter now.

Should I write tests for this kind of change @noracato?

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -33,16 +33,12 @@ $(document).on("turbo:load", function () {
       viewportMargin: Infinity,
     });
 
-    // Ctrl+Enter submit on the GQL editor.
-    if ($(this).attr("name") == "query") {
-      cm.on("keydown", function (el, event) {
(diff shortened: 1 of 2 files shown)


noracato (approved):
I brought back the immediate focus in the sandbox, because I know our modellers love that feature. For the rest, nice and simple solution!

mabijkerk:
🙌

PR #1443, opened 2024-06-26
Add constraint for hybrid offshore converter

noracato:
Make sure the converter (electrolyser) does not bid on the market when the cable is not big enough!


kndehaan (approved):
LGTM

kaskranenburgQ (commented):
LGTM!

PR #1450, opened 2024-08-19
CTM-ETM coupling migration of old refinery transformation inputs

mabijkerk:
To improve the coupling between the ETM and the CTM, new generic transformation nodes have been added for `chemical_other`, `chemical_refineries` and `chemical_fertilizers`. The generic transformation node for the refinery sector can transform crude oil to oil products, much like the existing refinery transformation nodes.

We have decided to migrate the external coupling transformation flows for refineries from the existing node to the new generic node. This PR handles that migration. The migration has not yet been run, so Semaphore fails on db schema.

To do for:

- [ ] Check the migration and improve if necessary @noracato 
- [ ] The dataset values in line 63 do not work and likely require a JSON that holds the present year demand of the `industry_refinery_transformation_crude_oil` or the  `industry_useful_demand_for_chemical_refineries_crude_oil_non_energetic` node @noracato 
- [ ] Depending on the way the JSON is set up, perhaps we need to convert the data from the JSON from MJ to PJ
- [ ] Check and run the migration @mabijkerk


mabijkerk:
@noracato I made some minor changes, most notably dividing the present demand by 1000.0 to get from TJ to PJ. Two questions:

1. What does the .gitkeep file do?
2. The not_defined input and output does not seem to be work.

The not_defined input should be set to 0.0 which should be simple enough to add I think? The not_defined output however should be allocated the refinery gas share. This is not the case now. In the example below, ~4% of output (refinery gas) is missing:

Screenshot 2024-08-22 at 09 03 50

db/migrate/20240816114754_crude_oil_transformation.rb
```diff
@@ -0,0 +1,78 @@
...
+      scenario.user_values[NEW_OUTPUT_SHARE.call(carrier)] = 0.0
+    end
+
+    EXISTING_CARRIERS.each do |carrier|
+      scenario.user_values[NEW_OUTPUT_SHARE.call(carrier)] =
+        scenario.user_values.delete(OLD_OUTPUT_SHARE.call(carrier)) || 0.0
+    end
+
+    scenario.user_values[NEW_OUTPUT_SHARE.call('not_defined')] =
+      scenario.user_values[NEW_OUTPUT_SHARE.call('refinery_gas')]
```
noracato, line 76:
```suggestion
      scenario.user_values.delete(OLD_OUTPUT_SHARE.call('refinery_gas')) || 0.0
```

noracato (commented):
This should be it!

PR #1452, opened 2024-08-22
Migrate external coupling inputs for coal gas distribution

mabijkerk:
In https://github.com/quintel/etsource/commit/3ae7109012b11fedd9e420111f1e2095f6bb7af2 the external coupling of coal gas distribution outputs has been updated. Most notably, a share group is added and the group is completed by the addition of an input for the output share going to final demand.

This migration sets the value for the new input for external coupling of coal gas distribution to final demand.


noracato (approved):
LGTM!

PR #1453, opened 2024-08-22
EV Migration

louispt1:
The migration renames the previous EV share keys. If there are no EV custom curves uploaded, it keeps the shares as is. Otherwise, it aggregates the uploaded curves based on their shares, removes the uploaded curves and creates a new attachment assigned to the custom profile. This profile is assigned a 100% share.

I do feel as though this could potentially be simplified.

Note: Not sure if I should remove the changes to the schema that happened during testing from this PR?


db/migrate/20240729080631_electric_vehicle_profiles.rb
```diff
@@ -0,0 +1,174 @@
...
+
+    # Set up the custom profile in the scenario.
+    setup_custom_profile(scenario, custom_profile)
+  end
+
+  # Renames keys in the scenario's user values based on the provided mapping.
+  def rename_keys(scenario, new_mapping)
+    new_mapping.each do |old_key, new_key|
+      # Transfer values from old keys to new keys, defaulting to 0 if the old key is not present.
+      scenario.user_values[new_key] = scenario.user_values.delete(old_key) || 0.0
```
noracato, line 108:
```suggestion
      # Transfer values from old keys to new keys, if the old key is present.
      next unless scenario.user_values.key? (old_key)

      scenario.user_values[new_key] = scenario.user_values.delete(old_key)
```

If none of the sliders in the group was touched, and now they all default to 0.0, the group does not sum too 100% and the scenario will break. It's best to just skip when the key was not there.

noracato (commented):
I'm leaving these comments for you @louispt1 to see why I changed some things. 
I'll commit my fixes afterwards. 

Using the attachment service was a great choice. Overall you did well, just a few things surrounding the share groups, and using Merit to mix curves as it has high performance and precision

kaskranenburgQ (commented):
Looks good!

PR #1455, opened 2024-08-23
CTM-ETM coupling: generic transformation nodes

noracato:
This PR allows for circularities in the graph via transformation nodes, and solves these circularities by locally setting net-shares and net-demands on the transformation nodes.

Algorithm of nettifying will be improved in the future.

Dependent on quintel/atlas#170, quintel/refinery#62


app/models/qernel/reconciliation/helper.rb
```diff
@@ -28,6 +28,8 @@ def supply_demand_balance(plugin)
           .installed_adapters_of_type(:producer)
           .sum(&:carrier_demand)
 
+        #  TODO: add transformation here
```
mabijkerk, line 31: @noracato there's an open to do here. Should it be addressed?

app/models/qernel/merit_facade/transformation_adapter.rb
```diff
@@ -0,0 +1,81 @@
...
+          number_of_units: target_api.number_of_units
+        )
+      end
+
+      def installed?
+        producer? || consumer?
+      end
+
+      # Does not inject demand, as this node is expected to be a preset demand node
+      # TODO: check if it should inject for carrier specfic
```
noracato, line 55: This one should also go indeed

app/models/qernel/reconciliation/helper.rb
noracato: Yep

app/serializers/reconciliation_csv_serializer.rb
```diff
@@ -5,14 +5,14 @@ class ReconciliationCSVSerializer < CausalityCurvesCSVSerializer
   private
 
   def producer_types
-    %i[producer import storage]
+    %i[producer flex]
   end
 
   def consumer_types
-    %i[consumer export storage]
+    %i[consumer flex]
```
mabijkerk, line 12: Why are export and storage removed here? These attributes still exist on ETSource for the natural gas nodes. For example `energy_export_natural_gas_backup` has the following:
```
- network_gas.profile = network_gas_export
```

louispt1 (commented):
Looks good, just spotted one TODO other the one Mathijs caught.

mabijkerk:
See https://github.com/quintel/etsource/pull/3107 for more information.

app/serializers/reconciliation_csv_serializer.rb
noracato: @kndehaan can you maybe check this question, as you were the one working on this?

kndehaan: I thought that this class was only used for the hydrogen csv, but now see that it's also used for the network gas csv, my bad. Changes reverted and class description updated with https://github.com/quintel/etengine/pull/1455/commits/be650f4e7da43326860f92cad370b430759c90bf.

PR #1457, opened 2024-08-23
CTM-ETM coupling: external input groups

noracato:
This PR introduces active couplings as part of a scenario, and inputs that can be disabled or enabled by coupling groups.

Two new API endpoint:
- uncouple: to uncouple one or more groups softly, or with the `force` parameter resets all coupling inputs like we used to do for uncoupling
- couple: activate one or more coupling groups


mabijkerk:
See https://github.com/quintel/etmodel/pull/4323 for more information.

louispt1 (commented):
LGTM

PR #1463, opened 2024-10-15
Download inputs as a csv 

### LINKED ISSUE #1449: Add data-export for inputs
We have a number of data-exports in our front-end relating to the output of a scenario. I would like to add a data-export for the inputs that have been set in a scenario:

- key
- min
- max
- present value
- future value
- unit
- share group

A lot of this information is already present in the inputs API, judging by our [docs](https://docs.energytransitionmodel.com/api/inputs). @kaskranenburgQ and @louispt1 can pick this up at some point.

louispt1:
Download inputs as a csv with:
- key
- min
- max
- present value: default
- future value: user
- unit
- share group 

at /api/v3/scenarios/2601971/inputs.csv

Closes #1449

### DIFF
--- a/app/controllers/api/v3/inputs_controller.rb
+++ b/app/controllers/api/v3/inputs_controller.rb
@@ -1,27 +1,31 @@
 # frozen_string_literal: true
+require 'csv'
 
 module Api
   module V3
     class InputsController < ::Api::V3::BaseController
       before_action do
         @scenario = Scenario.find(params[:scenario_id])
-        authorize!(:read, @scenario)
+        if user_signed_in? && current_user.can?(:read, @scenario)
+          authorize!(:read, @scenario)
+        end
       end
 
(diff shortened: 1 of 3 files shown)


mabijkerk:
Fast work! Two questions:

1. Normally when we download a data-export, the SessionID is appended to the file. Instead of inputs.csv I would then get inputs.123456.csv. Is this something we can add here as well?
2. When I go to the JSON inputs endpoint, all the default values are filled in. For example costs_bio_ethanol has "default": 0.948245519132. When I download the CSV for the same scenario however, this key (and a lot of other keys) have no default value. Is something going wrong here?

louispt1:
Good points! I have resolved both I think, and added some additional spec for the csv download - could you verify again @mabijkerk?

mabijkerk:
No I get this error I think:

Screenshot 2024-10-16 at 09 02 11

PR #1469, opened 2024-11-12
Migration building stock data of EU datasets

kndehaan:
This PR consists of a migration related to the update of building stock data for EU datasets. With this migration, the set values and default value settings of building stock inputs are preserved for existing scenarios.

Goes with:
- https://github.com/quintel/etsource/pull/3139
- https://github.com/quintel/etlocal/pull/560
- https://github.com/quintel/etdataset/pull/1023
- https://github.com/quintel/etengine/pull/1469


kaskranenburgQ:
I have tested the migration and it works!
There is one issue that arises:
If before the migration I have set the slider to a value that is higher than the new max value, and then run the migration I will see the new max values in the front-end:
Screenshot 2024-11-13 at 14 01 04

While actually in the engine, the slider value is equal to the old value I put in before the migration
Screenshot 2024-11-13 at 14 01 49

@noracato Is this behaviour expected?
This might confuse users.

noracato (commented):
Looks good!

Two things:

1. Let's remove the verbose output as it will clutter the logs on the server
2. I'm missing db/schema

noracato (approved):

db/migrate/20241112093940_set_eu_building_stock_values.rb
```diff
@@ -0,0 +1,153 @@
...
+      value = scenario_defaults[area]
+      if value.nil?
+        next
+      end
+      # If slider not set, set slider to the dataset default value
+      unless scenario.user_values.key?(key)
+        scenario.user_values[key] = value
+      end
+    end
+  end
```
mabijkerk, line 123: Isn't this part redundant? If the slider isn't set originally, then the slider just has the start value. If we then update the building stock, the start value changes and we don't have to set it explicitly right?

PR #1480, opened 2024-12-20
Updated Transition paths and saved scenarios to use the idp_client an…

louispt1:
Updated Transition paths and saved scenarios to use the idp_client and updated their spec to match. Minor changes to other pending specs to get them going again in the new authorization setup

Also:
- Refactored transition paths to be collections
- Refactored stubbing tokens to match the expected format


app/services/upsert_saved_scenario.rb
louispt1, line 1: @noracato does it work to just put the version in the params of the upsert method like this? If so then I think the engine APIs are good for now

noracato:
I feel like this is merged already?

PR #1481, opened 2024-12-20
Implemented version dropdown as in ETModel

louispt1:
Basically the same setup as ETModel but for Engine.

Please note: I tried to implement the logic for 'local' in the MyETM PR, but I could see that the 'Local' tag is coming through as a version option. I didn't have time to resolve that.


app/models/etengine/version.rb
```diff
@@ -0,0 +1,26 @@
...
+    # so that we know the exact date. This should be the day on which the minor version was released,
+    # not incremented for each small change.
+    DAY = 5
+
+    STRING = format('%<major>d.%<minor>02d', major: MAJOR, minor: MINOR)
+    DATE = Date.new(MAJOR, MINOR, DAY)
+
+    def self.prerelease?
+      DATE > Time.now.utc
+    end
```
noracato, line 24: We don't need this I think, only the tag.

app/views/layouts/application.html.haml
```diff
@@ -32,34 +32,47 @@
...
-                %li= link_to 'ETSource',  etsource_commits_path
+                %li.dropdown
+                  %a#version-button.nav-link.dropdown-toggle{ href: '#', data: { toggle: 'dropdown' }, role: 'button', aria: { haspopup: 'true', expanded: 'false' } }
+                    %span= "##{ETEngine::Version::TAG}"
+                    %b.caret
+                  %ul.dropdown-menu.version-dropdown.development{ aria: { labelledby: 'version-button' } }
+                    %li{ style: 'margin-bottom:1rem; margin-left:1rem;' }= t('versions.switch')
+                    - MyEtm::Version.all_other_versions.each do |version|
+                      %li
+                        = link_to "##{version.tag}"
```
noracato, line 43: link to needs a path

PR #1488, opened 2025-01-13
Adding migration for the removal of industry P2H WTP inputs

kaskranenburgQ:
This PR adds the migration that is necessary for the removal of industry wtp p2h inputs.
It is written to check whether one of the removed inputs is set, if this is the case, the regular input is set to the external coupling input. 

Goes together with:
https://github.com/quintel/etsource/pull/3185
https://github.com/quintel/etmodel/pull/4391

### DIFF
--- a/db/migrate/20250113141155_remove_external_wtp.rb
+++ b/db/migrate/20250113141155_remove_external_wtp.rb
@@ -12,13 +12,13 @@ class RemoveExternalWtp < ActiveRecord::Migration[7.0]
     external_coupling_industry_other_paper_p2h_wtp
   ].freeze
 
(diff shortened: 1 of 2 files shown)


mabijkerk:
I tested a scenario where I set all the external coupling WTP inputs and then migrated. This worked as expected. @louispt1 could you have a short look at the code to see if the code is structured as expected?

louispt1 (approved):
LGTM

kaskranenburgQ:
I added an extra function that removes the external wtp sliders from the scenario's after the input value migration. 
@louispt1 Could you have another look to check if this is implemented correctly?

PR #1490, opened 2025-01-22
Cleaning up

louispt1:
Same as ETModel, tried to resolve pending tests and clean up TODOs as well as fixing minor code smells and reorganising the routes.

Some things were just removed as they have moved into MyETM or are no longer necessary.


config/routes.rb
noracato, line 1: So much cleaner 🙂

app/controllers/api/v3/collections_controller.rb
```diff
@@ -27,20 +28,20 @@ def index
...
         render_not_found
       end
 
       def create
         UpsertTransitionPath.new(
-          endpoint_path: '/api/v1/transition_paths',
+          endpoint_path: '/api/v1/collections',
           method: :post
         ).call(
           params: params.permit!.to_h,
```
noracato, line 42:
```suggestion
          params: params.permit!.to_h.merge(version: Settings.version_tag),
```
Communicate the version on create

noracato (commented):
So nice and clean! Few points about communicating the version to MyETM, and backwards compatibility. Overall very happy to see it so neat, thank you!

noracato (approved):

PR #1493, opened 2025-01-30
Updated authenticate request flow to authenticate both PATs and JWTs

louispt1:
Basically now when the engine receives a JWT it will authorise it directly, and if it receives a PAT it will exchange the PAT for a JWT with MyETM.

For now this works, but in the future it would be nice to convert all tokens to JWT format and compress the amount of information in them so that they can be a reasonable size - that way we avoid these complicated authentication and authorization flows.


louispt1:
This may be unnecessary with the new identity flow

PR #1495, opened 2025-02-17
Use http client from Identity gem

noracato:
Bumps the identity gem and uses the http client from there to interact with MyETM.
Removes redundant auth logic already picked up by cancancan and identity.

Still open:
- Support audience claim as array (possible `include` instead of `==`) in `ETEngine::TokenDecoder`

Goes with:
quintel/etmodel#4408
quintel/my-etm#72
quintel/identity_rails#6


louispt1 (approved):
Looks good! A lot cleaner here already

PR #1498, opened 2025-02-19
Updated user controller to accept private_scenarios as a param

louispt1:
Updated user controller to accept private_scenarios as a param - I also adjusted it so it relies on the authorization of the BaseController

### DIFF
--- a/app/controllers/api/v3/users_controller.rb
+++ b/app/controllers/api/v3/users_controller.rb
@@ -3,11 +3,10 @@
 module Api
   module V3
     # Updates user information.
-    class UserController < BaseController
-      before_action :authorize_user!, only: :update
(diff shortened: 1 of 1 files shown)


app/controllers/api/v3/users_controller.rb
```diff
@@ -21,12 +20,12 @@ def destroy
...
-      def user
-        @user ||= current_user
+      def user_params
+        params.require(:user).permit(:name, :private_scenarios)
       end
 
-      def authorize_user!
-        head(:forbidden) if user.id != params.require(:id).to_i
+      def user
+        @user ||= current_user
```
noracato, line 28: Smart to use the built-in auth on current user!

noracato (approved):

PR #1503, opened 2025-03-05
Removed users from scenario serialisation

louispt1:
Removed users from the scenarios serialisation and updated spec to check user scenario associations without relying on the response.


noracato (approved):
Thanks @louispt1!

PR #1504, opened 2025-03-06
Use email from JWT to initialize and couple users

noracato:
Adds the email field back to the User in the db. We need the email to verify existing 'dirty' ScenarioUsers, and at this point `identity_user` is not yet available.

This transaction was failing before because it was matching on empty email fields when trying to couple users, rolling back the full user create each time.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -9,7 +9,7 @@ class User < ApplicationRecord
 
   attr_accessor :identity_user
 
-  delegate :email, :roles, :admin?, to: :identity_user, allow_nil: true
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -22,18 +22,27 @@ class User < ApplicationRecord
 
   after_create :couple_scenario_users
 
+  # If users are initialized from JWT, then their email is not available
```
louispt1, line 25: Ah I didn't realise! Nice one

louispt1 (approved):
Nice one!

PR #1505, opened 2025-03-06
Rescue JSON::JWT exceptions as well as decode errors

louispt1:
This way users should receive some output when they use an old or malformed token alerting them to the fact that the token is the issue.


noracato (approved):
Brilliant

PR #1506, opened 2025-03-07
Functional inputs serialisation to csv

louispt1:
Goes with this PR from ETModel: https://github.com/quintel/etmodel/pull/4433


mabijkerk (changes requested):
@noracato should review the code itself, but I tested it and it works as I would expect 🙌🏻 Only question @louispt1 is why you removed one of the migrations? Finally, a next step would be to add the locales to this download, but that would be an enhancement. Let's first get this PR merged.

louispt1:
This has been picked up with a different approach in https://github.com/quintel/etmodel/pull/4433

PR #1510, opened 2025-03-13
Adjust engine db importer for tar.xz

louispt1:
Adjust the engine db importer to handle differrent file types correctly based on the current setup.

I tested this locally and it works. Sometimes the .xz gets stripped from the filename by the time it gets to slack or into your downloads, so this accounts for that possibility and handles incorrectly parsing shell keywords.


noracato (approved):
LGTM! Thanks!

PR #1513, opened 2025-03-20
Fix for tokens on beta

louispt1:
Adjust method to strip prefix for tokens as remove was greedily removing etm_ and leaving the beta behind.


noracato (approved):
Thanks @louispt1

Issue #1524, opened 2025-03-24
Fix the JSON response of the custom curves controller by updating the `CustomCurveSerializer`


noracato:
The JSON requests for the show endpoint serialises the contents of the attachments using the `CustomCurveSerializer`. So, instead of the csv response that just dumps the file, the JSON response will parse the content.

We should rename the `attachment` here to `user_curve` and replace references to the file contents (`@custom_curve`) with the `user_curve.curve`.

Issue #1522, opened 2025-03-24
Adjust `CurveHandler::DetachService`


noracato:
We used the attach and detach services to attach and remove the files from the attachment. Now that we will not longer store the files directly, we should rewrite these modules according to our new needs.

The attach service should read the file contents of the uploaded csv, and convert it to a `UserCurve` and save it.

The detach service can simply remove the record.

Issue #1520, opened 2025-03-24
Research if we want move from type `Array` directly to `Merit::Curve` with a custom serializer based on MessagePack.


noracato:
It can be very beneficial to directly deserialize a saved curve to a `Merit::Curve` as we are always doing it when grabbing the curve. This will save a lot of extra code and later on, but will force us to create a custom serializer/deserializer. 

We need to do one or two hours of quick research on how to create such a serializer using MessagePack, and then create a small dev plan.

Issue #1518, opened 2025-03-24
Update `CustomCurveCollection`


noracato:
The graph (or actually, the dataset behind the graph) will be given a `CustomCurveCollection.` It's in the name: a collection of `Merit::Curve` objects with as a key the custom curve they represent, as attached to the current scenario.

We need to update this class to no longer read csvs and convert them in `Merit::Curves` but to simply grab the curve from the `UserCurves `and either

 -   if we chose to go for the recommendation in 1 (see parent issue), we don't need to do anything here
 -   if we did not, convert the array into a `Merit::Curve`

PR #1530, opened 2025-03-27
Attach / Detach user curves

### LINKED ISSUE #1521: Adjust `CurveHandler::AttachService`

louispt1:
This PR builds upon #1529 to refactor the attach and detach services and their spec to be able work with the new UserCurve model. Notably - adjusting attach and detach causes the custom curves endpoint to break, which is reflected in the spec (currently failing for that endpoint).

Closes #1521, #1522, #1519, #1518, #1520, #1523, #1524

Edit: This PR supersedes #1529.

### DIFF
--- a/app/controllers/api/v3/custom_curves_controller.rb
+++ b/app/controllers/api/v3/custom_curves_controller.rb
@@ -1,10 +1,11 @@
+# frozen_string_literal: true
+
 module Api
   module V3
     # Provides the ability to upload or remove a custom curve from a scenario.
     #
-    # Currently only the imported electricity price curve can be changed, but this controller is
-    # intended to allow additional custom curves to be added later without requiring changes to the
-    # REST API.
+    # This controller now uses the UserCurve model, which stores curves in the database using
+    # MessagePack-encoded Merit::Curve objects.
     class CustomCurvesController < BaseController
       include ActionController::MimeResponds
       include UsesScenario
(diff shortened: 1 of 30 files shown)


noracato (commented):
Amazing work! Very well done!

I'll push a commit with some minor database optimisations. That will hopefully make loading a scenario even faster than what you did here!

Maybe we can slightly rethink the serializer on the model. For instance keeping the data as an array with Rails built in serialising and just adding a merit_curve method on the model that will cast it for us.
What do you think? Or is storing it like this already quite optimised? Would love to hear your thoughts.

louispt1:
I think the main trade-off is performance vs readability/maintainability. My feeling is that this is quite optimised, but I suppose it also wouldn't be too difficult to directly test the differences and then decide? If the array approach is only a little bit slower (etc) then it might be preferable from the perspective of keeping things Rails-y.

Let me know if you think it's worth testing them against each other

noracato:
We could also just merge it like this and open an issue for possible optimisation for the future. I quite like it like this as well.

Last thing to do is to write the big migration that will take all the existing csvs and converts them into the new records.

louispt1:
I committed the migration - it takes a long time to run, but worked for me locally at least. There is the question of whether we should 'touch' the scenarios that have been migrated, or just use `raise_if_no_changes: false`. At the moment I have both.

I also added the curve methods back to the scenario attachment model in this commit, with a TODO to remove them once we've run the migration.

noracato (approved):
Time to rebase and merge 🎉 

Awesome work

PR #1531, opened 2025-03-27
Bump Merit to quintel/merit@5ac78a1

### LINKED ISSUE #1482: Price-setting issues with forecasting algorithm
This issue is a summary and conclusions of issue #1403.

There are 2 separate issues we have found due to the price-setting algorithm:

Issue 1: if forecasting is on and there is curtailment, during battery discharge, the price of the cheapest dispatchable is taken. This can be seen in the following scenario: https://energytransitionmodel.com/saved_scenarios/19128
Hypothesis: battery is acting as must-run producer, but in pricesetting apparently merit still thinks price should be set by dispatchables, leading to the hypothesis that pricesetting does not see it as a must-run producer (when curtailment is happening) (edited) 

Issue 2: if forecasting is on, during charge, the highest willingness to accept (of other batteries or of DSR) is taken. This behavior (together with the behavior of issue 1) can be seen in the following scenario:  https://energytransitionmodel.com/saved_scenarios/19140
Hypothesis: this seems to be the exact inverse of issue1. Instead of must-run producing and taking the price of the cheapest dispatchable, it is must-run consumption and taking the price of the most expensive flexible consumer. 

Long-term considetation: Should we rewrite merit?

@noracato Could you try to find the cause of this behaviour in the price-setting at one point?

Notifying @mabijkerk

noracato:
Closes #1482
Goes with quintel/merit#156

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,7 +76,7 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
-gem 'quintel_merit', ref: '76f2343', github: 'quintel/merit'
+gem 'quintel_merit', ref: '7b7a3cf', github: 'quintel/merit'
 gem 'atlas',         ref: '64bb548', github: 'quintel/atlas'
 gem 'fever',         ref: '2a91194', github: 'quintel/fever'
 gem 'refinery',      ref: 'c39c9b1', github: 'quintel/refinery'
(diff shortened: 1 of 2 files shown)


noracato:
@kaskranenburgQ could you review if the results are correct? Meaning: is the behaviour of the price curve as expected form our discussions in the issue?

kaskranenburgQ (approved):
LGTM!
Tested various scenarios including scenarios that where affected by the bug. 
These are fixed now.

PR #1528, opened 2025-03-24
Update to Rails 7.1 with messagepack as base serialiser

noracato:
Update Rails to 7.1.5 in order to prepare for issue #1509

PR #1532, opened 2025-03-28
Update to ruby 3.2 and rails 7.1 with messagepack as default message serializer

### LINKED ISSUE #1515: Update DeepClone (gem we use has last been updated 8 years ago)
We do an order of things to optimise the graph calulations. One of these is to cache a precalculated, optimised present graph for each dataset, as this graph does not tend to change (we don't allow many changes to the present).

When taken from the cache, this present graph is copied to form a base for the future graph, and the rest of the 'future' calculations will be done on this copy.

It's important that when we copy the present graph to the initial future graph, we make sure to _really_ copy everything. Many programming languages like to favor shallow copies (where some inner items still refer to the same object) over deep copying (each item and all inner items will become new objects), to save space and time. When copying the graph we explicitly need a deep copy, in order to not accidentally alter the present graph when updating something in the future graph. 

In the [etsource loader](https://github.com/quintel/etengine/blob/d08e25f2e09eb5b88ceba20602618c4877ab44de/app/models/etsource/loader.rb#L19), where the caching and copying happens, we use `DeepClone`, which comes from the gem `ruby_deep_clone` (last release 8 years ago, last supported ruby version 2.4).

Currently, we cannot upgrade ETEngine from Ruby 3.1 (which reaches end of life this April) to Ruby 3.2 because the deep clone gem causes a segmentation fault. 

Luckily there seem to be many other deep cloning gems in the market. We should check them out and pick a new one - or research if Rails now comes shipped with something similar out of the box. This could also potentially make the engine a bit faster, since new implementations would make use of new Ruby features.

### LINKED ISSUE #1525: Move to ruby 3.2
As ruby 3.1 will reach end of support soon, and we have been getting tons of deprecation warnings. We should update to ruby 3.2. A start has been made on `ruby-3.2`, but some gems should be replaced (see for instance #1515).

This branch is based on a point where the Rails version has already been updated to 7.1.

noracato:
Supersedes #1528 

DeepClone has been replaced by a simple Marshal approach for now. It could be a great next step to write custom (messagepack) serialisation and deseralisation later to optimise the cloning procedures, as Marshal is known to be one of more slow approaches.

Closes #1515 
Closes #1525


louispt1 (approved):
LGTM - I guess Capistrano is no longer used anywhere? Also good spot removing some of those auth checks that are no longer necessary.

for narray and gctools - I guess we just don't use those anymore? Seems like some sort of gc tracing tool could be useful, although maybe that's fully handled by sentry these days

noracato:
Capistrano was our old deploy gem, before we did CD with Semaphore.

As for gctools, they are indeed to track garbage collection. I could not update to something that was still supported, and I couldn't see them being used anywhere - and now that we are profiling with Sentry we get insights enough indeed!

narray is still in there, it's a dependency of Merit and should be updated via that gem. Pinning it in the engine gave me some trouble

PR #1536, opened 2025-04-03
Expect or convert to array in savedscenario response.

louispt1:
We can look at adding pagination back in the future. Maybe it's overkill to check the format both in index and hydrate?

Also, this is probably a separate issue but:
```
      before_action(only: %i[index show]) do
        authorize!(:read, @scenario)
      end

      before_action(only: %i[create update]) do
        authorize!(:write, @scenario)
      end
```

the @scenario won't be set when interacting with this endpoint via the api right? So for anything other than view actions I think there'll be a 403 error.


noracato (approved):
I think this is great for now! The authorization should be a separate issue indeed

PR #1539, opened 2025-04-09
Serialize user values

louispt1:
This PR introduces msgpack serialization for user values on scenarios, and a migration for the existing scenarios. Some adjustments to the spec were also necessary.

Closes [#1538](https://github.com/quintel/etengine/issues/1538)


noracato (approved):
Fixed on last thing in the update method in the migration.

Let's merge 🎉 

Nicely done with the temporary model

PR #1543, opened 2025-04-10
Serialize balanced values with MessagePack

louispt1:
Following the pattern of user_values, worked locally.

Closes: #1540

### DIFF
--- a/db/migrate/20250409134358_serialize_balanced_values.rb
+++ b/db/migrate/20250409134358_serialize_balanced_values.rb
@@ -1,0 +1,34 @@
+class SerializeBalancedValues < ActiveRecord::Migration[7.1]
+  include ETEngine::ScenarioMigration
+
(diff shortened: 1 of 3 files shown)


db/migrate/20250409134358_serialize_balanced_values.rb
```diff
@@ -0,0 +1,34 @@
...
+  # Temporary model
+  class ScenarioForMigration < ActiveRecord::Base
+    self.table_name = 'scenarios'
+  end
+
+  require 'msgpack'
+
+  def up
+    change_column :scenarios, :balanced_values, :text, size: :medium
+    add_column :scenarios, :balanced_values_binary, :binary, limit: 16.megabytes
```
noracato, line 13: I feel like we don't need such a big size for the balanced values. Medium should be sufficient. Or maybe even small...?

PR #1544, opened 2025-04-10
Serialize active couplings with MessagePack

louispt1:
I couldn't quite get this to work! When I run the debugger and investigate the `msgpack_blob` I get good MessagePack encoded data, but in the console it comes through as an empty array (see below). The spec failures are either from converting from a symbol to a string or from the empty array issue.

```
[2] pry(main)> s.active_couplings
=> []
[3] pry(main)> s.active_couplings_old
=> "---\n- :external_model_industry\n- :industry_chemical_other\n- :industry_metal_steel\n"
```

### DIFF
--- a/app/models/scenario.rb
+++ b/app/models/scenario.rb
@@ -15,7 +15,7 @@ class Scenario < ApplicationRecord
 
   serialize  :user_values, type: Hash, coder: MessagePack
   store      :balanced_values
-  serialize  :active_couplings, type: Array
+  serialize  :active_couplings, type: Array, coder: MessagePack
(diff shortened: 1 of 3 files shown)


louispt1:
I may have another solution for this so we can avoid using a custom wrapper - for the metadata I added this initialiser:

```
require 'msgpack'

module MessagePack
  class << self
    alias_method :load, :unpack
  end
end
```

because ActiveRecord otherwises confuses the load and unpack methods. We could see if that works here as well.

noracato (approved):

Issue #1485, opened 2025-01-09
Network gas hourly balance should treat export of green and natural gas equally


mabijkerk:
**Issue**
There is a difference between the annual network gas balance and the hourly. See full background in https://github.com/quintel/etsource/issues/3080. Part of the difference is caused by the different type for green gas and natural gas export.

There are four types of export for network gas, which have the following
| Key                               | Network Gas Type     |
|-----------------------------------|----------------------|
| energy_export_greengas_backup     | consumer             |
| energy_export_greengas_baseload   | consumer             |
| energy_export_natural_gas_backup  | export               |
| energy_export_natural_gas_baseload| consumer             |

It is clear that only the backup export for natural gas has the type `export`. This is strange because, if the mix would be set 100% to green gas, then we should expect the balancing for backup export to work exactly the same.

**Solution**
I have not actually verified that this issue causes an imbalance, but based on the code it is likely the case. It however only occurs when there is oversupply of green gas. Because green gas supply is demand-driven by default, this issue will only arise when the model is externally coupled and green gas also can be produced from the transformation nodes. This therefore does not have top priority.

- [ ] Check if the issue occurs
- [ ] Discuss how the balancing of green and natural gas backup export should be algined
- [ ] Implement discussed solution

Issue #1483, opened 2025-01-09
Gas power fuelmix should only count for share of network gas in hourly balance


mabijkerk:
**Issue**
There is a difference between the annual network gas balance and the hourly. See full background in https://github.com/quintel/etsource/issues/3080. Part of the difference is caused by the way that gas power fuelmix is handled.

```
EACH(
  V(energy_import_natural_gas,output_of_natural_gas),
  SUM(V(energy_import_natural_gas,network_gas_output_curve))*MJ_PER_MWH
)
```
First with a completely blank scenario:
```
[
  493,239,088,457.2776,
  520,899,250,749.96313,
]
```
Then a scenario with the share of network gas in the gas power fuelmix set to 100%:
```
[
  520,372,089,527.83545,
  520,372,089,527.84467,
]
```

This is because nodes that participate in the network gas balancing can have a demand carrier assigned. For the industry CHPs that use the gas power fuelmix, it looks like this:
```
- network_gas.demand_carrier = gas_power_fuelmix
- network_gas.profile = self: electricity_output_curve
- network_gas.type = consumer
```

In the [consumer_adapter](https://github.com/quintel/etengine/blob/master/app/models/qernel/reconciliation/consumer_adapter.rb#L30), there is no special case for gas_power_fuelmix. This indicates that it is treated as full network gas input in the hourly calculations, even though it can be part bio-oil or oil in the annual energy graph.

**Solution**
There are multiple potential solutions. We need to have a brainstorm to determine the robustness of each solution and the required work.

1. Use some kind of recursive factor to determine the share of network gas in gas power fuelmix
2. Remove the gas power fuel mixer and directly input network gas, bio-oil and oil on the CHP nodes
3. Specify the network gas demand on the gas power fuel mixer, but use some kind of subordinate adapter to allocate it to the specific CHP nodes
4. ...?

mabijkerk:
This issue already apparently existed in https://github.com/quintel/etsource/issues/2285. Closing that issue and leaving this open.

louispt1:
# Description
I've looked into this issue, and I would like to propose adding network gas directly on the mixer: `energy_mixer_for_gas_power_fuel.converter.ad` and removing the network gas configurations on the CHP nodes. That way the CHP nodes don't participate in network gas reconciliation, and reconciliation can happen upstream at the mixer. In my opinion this makes more sense because then we reconcile when network gas actually enters the system, rather than downstream. This approach also has the benefit of being quite simple.

[...]

PR #1546, opened 2025-04-10
Reconciliation transformation nodes are part of the supply demand balance

### LINKED ISSUE #1484: Network gas hourly balance should include transformation
**Issue**
There is a difference between the annual network gas balance and the hourly. See full background in https://github.com/quintel/etsource/issues/3080. Part of the difference is caused by the way that the transformation nodes are handled.

Both the Reconciliation [export_adapter](https://github.com/quintel/etengine/blob/master/app/models/qernel/reconciliation/export_adapter.rb#L16) and [import_adapter](https://github.com/quintel/etengine/blob/master/app/models/qernel/reconciliation/import_adapter.rb) use the [helper](https://github.com/quintel/etengine/blob/master/app/models/qernel/reconciliation/helper.rb#L22-L32) function `supply_demand_balance`:

```ruby
      def supply_demand_balance(plugin)
        consumption = plugin
          .installed_adapters_of_type(:consumer)
          .sum(&:carrier_demand)

        production = plugin
          .installed_adapters_of_type(:producer)
          .sum(&:carrier_demand)

        production - consumption
      end
```

This balance only looks at consumer and producer types that participate in network gas. This used to make sense, because the only other types are import, export and storage. Import and export are set using this function, so need to balance out any difference between consumer and producer types. Storage only balances hourly differences and is therefore net zero over the year.

However, recently the transformation type has been added for the external coupling transformation nodes. These can be both producer and consumer. This is not included in the `supply_demand_balance` function.

**Solution**
I have not actually verified that this issue causes an imbalance, but based on the code it is likely the case. Since it only occurs when the model is externally coupled, this does not have top-priority.

- [x] Check if the issue occurs
- [ ] Check if it is caused by the helper function
- [ ] Include transformation supply in producer and transformation demand in consumer

noracato:
Fixes an issue where the transformation nodes where not taken into account when calculating the supply and demand balance in reconciliation.

Closes #1484

### DIFF
--- a/app/models/qernel/reconciliation/transformation_adapter.rb
+++ b/app/models/qernel/reconciliation/transformation_adapter.rb
@@ -21,12 +21,16 @@ module Qernel
         end
       end
 
+      def carrier_demand
+        carrier_demand_input - carrier_demand_output
+      end
+
       def carrier_demand_input
-        @carrier_demand_input || raise("carrier_demand_input not yet calulated for #{@node.key}")
+        @carrier_demand_input.presence || raise("carrier_demand_input not yet calulated for #{@node.key}")
       end
 
       def carrier_demand_output
-        @carrier_demand_output || raise("carrier_demand_output not yet calulated for #{@node.key}")
+        @carrier_demand_output.presence || raise("carrier_demand_output not yet calulated for #{@node.key}")
       end
 
       def demand_curve
--- a/app/models/qernel/reconciliation/transformation_adapter.rb
+++ b/app/models/qernel/reconciliation/transformation_adapter.rb
@@ -21,8 +21,13 @@ module Qernel
         end
       end
 
+      # Unused for real demand calculations.
+      #
+      # Should be positive for plugin.installed_adapters_of_type to verify that
+      # this adapter is active. Taking the max will ensure a positive demand if
+      # either consuming or producing
       def carrier_demand
-        carrier_demand_input - carrier_demand_output
+        [carrier_demand_input, carrier_demand_output].max
       end
 
       def carrier_demand_input
--- a/app/models/qernel/reconciliation/adapter.rb
+++ b/app/models/qernel/reconciliation/adapter.rb
@@ -40,7 +40,7 @@ module Qernel
       end
 
       def carrier_demand
-        @carrier_demand || raise("carrier_demand not yet calulated for #{@node.key}")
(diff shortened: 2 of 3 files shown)


app/models/qernel/reconciliation/transformation_adapter.rb
```diff
@@ -21,12 +21,16 @@ def setup(phase:)
         end
       end
 
+      def carrier_demand
+        carrier_demand_input - carrier_demand_output
+      end
+
       def carrier_demand_input
-        @carrier_demand_input || raise("carrier_demand_input not yet calulated for #{@node.key}")
+        @carrier_demand_input.presence || raise("carrier_demand_input not yet calulated for #{@node.key}")
```
louispt1, line 29: Why does this work?

louispt1 (approved):
neat solution! I'm still confused why it works - is it because it sometimes gets set to nil?

app/models/qernel/reconciliation/transformation_adapter.rb
noracato: It's because it gets sometimes set to zero for transformation (either input or output) and it will fall in to the other side of the OR.

app/models/qernel/reconciliation/transformation_adapter.rb
```diff
@@ -21,12 +21,16 @@ def setup(phase:)
         end
       end
 
+      def carrier_demand
+        carrier_demand_input - carrier_demand_output
+      end
+
```
noracato, line 27: @mabijkerk can you verify this? Or should this value always be the net difference (so with an `abs` around it)?

mabijkerk:
When I run the following query for an affected scenario...
```
EACH(
  SUM(V(energy_import_natural_gas,network_gas_output_curve))*MJ_PER_MWH,
  V(energy_import_natural_gas,output_of_natural_gas)
)
```
I get the following on pro:
```
[
  77,007,322,238.06111,
  167,117,371,893.3646,
]
```
With the rebased branch, I get the following locally:
```
[
  168,413,839,877.1492,
  168,461,328,492.99142,
]
```

This means that the fix does reduce the delta by a significant amount, but there still is a 0.05 PJ difference between the results. I think this difference probably does not come from the transformation node issue. There are some other known issues that may cause this, like https://github.com/quintel/etengine/issues/1485 or https://github.com/quintel/etengine/issues/1483.

mabijkerk (approved):

PR #1547, opened 2025-04-10
Msgpack curves and user values fallout

louispt1:
Clone user curves, added (very) basic persistable spec and updated migratable_since to match new user_values format


noracato (approved):
Looks great!

PR #1548, opened 2025-04-14
Updates the user curves migration not to overwrite the set inputs

noracato:
Updates the migration from attachments to user curves not to set inputs. Removes one of the duplicate migrations


louispt1 (approved):
LGTM! Thanks

PR #1552, opened 2025-04-15
Include transformation adapters in demand and supply curves in reconciliation

noracato:
Follow up of #1484 

We fixed including the transformation nodes in the yearly supply demand balance, but did not include them in the hourly curves.

Hereby.


mabijkerk (approved):

louispt1 (approved):
LGTM

PR #1553, opened 2025-04-17
Msgpack metadata

### LINKED ISSUE #1541: Msgpack: metadata

louispt1:
Serializes metadata with MessagePack, works locally. This uses the initialiser method - apparently Messagepack 2.0 would not require an alias, but because we use Bootsnap which requires MessagePack 1.* this is the simplest approach to patch the MessagePack methods.

Closes #1541

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,6 @@ gem 'json'
 
 gem 'rake'
 gem 'activeresource', '~> 6.0'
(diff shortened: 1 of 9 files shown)


noracato (approved):

noracato:
Rebase and merge after next deploy

PR #1555, opened 2025-04-17
Serialize user sortables with MessagePack

### LINKED ISSUE #1542: Msgpack: user_sortable

louispt1:
Serializes all four user sortable tables with MessagePack. Works locally.

Closes #1542

### DIFF
--- a/app/models/hydrogen_supply_order.rb
+++ b/app/models/hydrogen_supply_order.rb
@@ -5,7 +5,6 @@ class HydrogenSupplyOrder < ApplicationRecord
   include UserSortable
 
   validates :scenario_id, presence: true, uniqueness: true
-  serialize :order, type: Array, coder: MessagePack
 
   def self.default_order
     Etsource::Config.hydrogen_supply_order
--- a/app/models/hydrogen_supply_order.rb
+++ b/app/models/hydrogen_supply_order.rb
@@ -5,6 +5,7 @@ class HydrogenSupplyOrder < ApplicationRecord
   include UserSortable
 
   validates :scenario_id, presence: true, uniqueness: true
+  serialize :order, type: Array, coder: MessagePack
 
   def self.default_order
     Etsource::Config.hydrogen_supply_order
--- a/app/models/forecast_storage_order.rb
+++ b/app/models/forecast_storage_order.rb
@@ -5,6 +5,7 @@ class ForecastStorageOrder < ApplicationRecord
   include UserSortable
 
   validates :scenario_id, presence: true, uniqueness: true
+  serialize :order, type: Array, coder: MessagePack
 
   def self.default_order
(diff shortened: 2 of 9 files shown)


noracato (approved):
Somehow after running this migration on the db, the scenario changed a tiny bit and is now the exact same as pro.

This is II3050 v2 Nat 2050

### Local
Screenshot 2025-04-17 at 18 22 58

### Beta
Screenshot 2025-04-17 at 18 23 06

### Pro
Screenshot 2025-04-17 at 18 23 15

louispt1:
> Somehow after running this migration on the db, the scenario changed a tiny bit and is now the exact same as pro.
> 
> This is II3050 v2 Nat 2050
> 
> ### Local
> Screenshot 2025-04-17 at 18 22 58
> ### Beta
> Screenshot 2025-04-17 at 18 23 06
> ### Pro
> Screenshot 2025-04-17 at 18 23 15

Is this something we should investigate further? Were the differences between beta and pro a result of the user sortables?

Maybe we should try run it on a pro db and see if it changes the values

noracato:
I tried to test it with a pro database, but the pro database does not contain any of the II3050v2 scenarios. Let's wait until we can properly test before we merge.

noracato:
I believe the difference is because we forgot to include the three heat network orders (low temp, medium and high) in the migration. But I will keep on checking!

louispt1:
Added heat network orders and households_space_heating_producer_orders:

Local (pre migration):
image

Local (after migration):
image

Beta:
image

Pro:
image

So it seems to be consistent with the local version from pre-migration, the question is why are there differences between local and beta even before any changes are made? This must be an issue with the database dump. I'm now trying to test with  the single scenario dump method... will update here when I have.

louispt1:
👍 

Local (post-migration):
image

Beta:
image

Pro:
image

app/models/hydrogen_supply_order.rb
```diff
@@ -5,6 +5,7 @@ class HydrogenSupplyOrder < ApplicationRecord
   include UserSortable
 
   validates :scenario_id, presence: true, uniqueness: true
+  serialize :order, type: Array, coder: MessagePack
```
noracato, line 8:
```suggestion
```

Let's update the serilaisation in `UserSortable` instead of adding it to each sortable. I think it could become confusing when the serialisation is defined in two places. 

Also: if we change the serialisation in `UserSortable` we should port the heat network sortables as well, as they will inherit the new serialisation.

noracato (changes requested):

app/models/hydrogen_supply_order.rb
louispt1: @noracato I made the changes! Makes tons of sense. The heat network sortables are in the migration as well

noracato (approved):

PR #1563, opened 2025-04-24
User curve upload issue

### LINKED ISSUE #1559: Uploading FLH custom curves in correct format gives incorrect scenario results on beta
**Background**
There seems to be a problem when uploading custom curves with type `capacity_profile` on beta. These are the curves that should sum to the total full load hours (for example wind and solar profiles). 

When uploading wind and solar profiles in a scenario on beta in the `capacity_profile` format, it gives very strange results:

Image

The total full load hours in the user_curve inputs are however correct:

Image

Through this route, it is therefore not possible to replicate the scenario from live to beta. 

When uploading the wind and solar profiles as normalised curve (type `profile`), the values in user_curve are (obviously) incorrect since they sum to 0.00278:

Image

But when subsequently correcting the user_curve inputs manually to the correct values from the live scenario, I do get the same results as the scenario on live. Basically, that's because the shape of the normalised profile is correct and apparently scales with the provided full load hours for the user_curve inputs. 

**Problem**
It becomes clear that trying to copy a scenario from live to beta the proper way (specifically regarding uploading curves in the correct format) leads to very wrong results. 

To obtain the same scenario on beta currently requires you to upload a normalised profile (instead of capacity_profile) and then override the user value with the value from the scenario on live.

This is not the desired way to go and should be fixed. Moreover, it is definitely not desirable that this problem would pass on to the live server. Before the deploy, it should be properly tested if this issue could be fixed or at least will not occur on the live server. 

Notifying @mabijkerk @noracato @louispt1

louispt1:
The PR solves the issue of uploading curves either setting the flh inputs for user curves extremely low so that they approach 0, and solves the issue of uploaded curves either being double-normalised or not normalised at all.

The issue is elaborated in  #1559.

Closes #1559.


kndehaan:
It seems to give the correct scenario results now when uploading the `capacity_profile` via scenario-tools! Also, the correct FLHs are shown in the user curve input

image

image


Two peculiarities: 
- Full load hours in the Modify profiles sections are shown to be zero (in contrast to the flh user curve value)
Screenshot 2025-04-24 at 17 09 18
- Similarly, the full_load_hours parameter in the custom curve info via the engine shows 0.0002778 http://localhost:3000/api/v3/scenarios/xxx/custom_curves?include_internal=true
Screenshot 2025-04-24 at 17 11 12

Basically, the functionality seems to work as it should now, only deriving the FLH in the two locations mentioned above (front-end Modify profiles and engine overview of uploaded custom curves) goes wrong. Do you think it is easy to fix? @louispt1 

To me it doesn't seem necessary, but if it's very easy then why not..

louispt1:
Nice pickups @kndehaan! I've just pushed a fix to the FLH as well. It should be possible to download the curves as they were uploaded (not-normalised) too, but I think I'll pick that up as part of a separate issue. If you're happy with the outcomes then I will ask @noracato to do a final review!

kndehaan (approved):
Very good! All seems to work well now and the correct FLH now also show in the front-end.

kndehaan (changes requested):
Actually, I just encountered an issue related to retrieving custom curves on beta and locally. 

 When I use the API (via scenario-tools) for obtaining the curves on beta or locally, an error arises from the `get_custom_curves` function:

```
    float_values = [float(value) for value in decoded_response]
ValueError: could not convert string to float: ''
```
It seems that `decoded_response` has a length of 8761 values, from which the last is an empty string which I think causes the error. It seems like this can be fixed easily. 

This problem does not exist on the live server.

kndehaan (approved):
Works well now via scenario-tools i.c.w. the fix in https://github.com/quintel/scenario-tools/pull/46

noracato (commented):
Happy you managed to solve the issue!

A few comments: 
- We should make a spec for the new serializer
- Calling `Scenario.find` in the serializer triggers an extra call to the db: can we be smarter about this?
- If the endpoint starts to return 8671 values instead of 8760, that is a bug in the endpoint **not** in the scenario tools. Let's check that out!

Edit: I'll pick up the changes this morning!

noracato:
I pushed the changes! I improved the original capacity profile serialiser instead, to avoid confusion. 

@kndehaan would you mind giving it one final check before we merge?

louispt1:
> I pushed the changes! I improved the original capacity profile serialiser instead, to avoid confusion.
> 
I think this setup will help simplify the code in [this issue](https://github.com/quintel/etengine/pull/1565) too
 - thanks!

kndehaan:
> @kndehaan would you mind giving it one final check before we merge?

Works as it should, good to merge!

PR #1564, opened 2025-04-25
Add methods for peak and volume of blackout hours

louispt1:
Goes with https://github.com/quintel/merit/pull/157

This is a draft PR for now as I'd like to determine a test case and test the changes with the commit on Merit. Perhaps we can do that together @kaskranenburgQ?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -74,8 +74,8 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
-gem 'quintel_merit', ref: '7b7a3cf', github: 'quintel/merit'
-gem 'atlas',         ref: '64bb548', github: 'quintel/atlas'
+gem 'quintel_merit', ref: 'd79f11d', github: 'quintel/merit'
+gem 'atlas',         ref: '8e854b1', github: 'quintel/atlas'
 gem 'fever',         ref: '2a91194', github: 'quintel/fever'
 gem 'refinery',      ref: 'c39c9b1', github: 'quintel/refinery'
 gem 'rubel',         ref: 'e36554a', github: 'quintel/rubel'
--- a/app/models/qernel/graph_api/energy.rb
+++ b/app/models/qernel/graph_api/energy.rb
@@ -25,18 +25,20 @@ module Qernel
         end.flatten.compact.sum
       end
 
-      # Demand of electricity for all final demand nodes
+      # Demand of electricity for all final demand nodes (MWh)
       def final_demand_for_electricity
         group_demand_for_electricity(:final_demand_group)
       end
 
       # Demand of electricity for all nodes which do not belong to the final_demand_group but
-      # nevertheless consume electricity.
+      # nevertheless consume electricity. (MWh)
       def non_final_demand_for_electricity
         group_demand_for_electricity(:non_final_electricity_demand_nodes)
       end
 
       # Demand of electricity for all nodes which belong to the named group.
+      #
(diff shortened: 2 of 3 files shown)


noracato:
If you push you merit branch, you can specify that merit commit in the gemfile on this engine branch. That way it's easy for others to test!

louispt1:
> If you push you merit branch, you can specify that merit commit in the gemfile on this engine branch. That way it's easy for others to test!

Done! Good idea

kaskranenburgQ:
We can test this by using the export "Electricity load curves". 
In this export the deficit for each hour is given.
We can double check if all the input and output curves in the export add up with the deficits. 

I have added ETSource queries for the table in [this commit](https://github.com/quintel/etsource/commit/bdb59462e221aeddc58cb22dd18ece32dc11f410) on [this branch](https://github.com/quintel/etsource/tree/blackout). We can test the work with these queries.

kaskranenburgQ:
NTH: The engine methods are now named after the old term "blackout_hours", since we no longer use this term we should rename these to power_shortage_hours.

louispt1:
FYI - I have updated the names of the methods as per the NTH above. Notifying @kaskranenburgQ

noracato (approved):

Gemfile
```diff
@@ -74,7 +74,7 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
-gem 'quintel_merit', ref: '7b7a3cf', github: 'quintel/merit'
+gem 'quintel_merit', ref: 'bbce9b01788d0b696dd831dac3d1e5f675da5141', github: 'quintel/merit'
```
mabijkerk, line 77: Why use the full hash here?

app/models/qernel/graph_api/energy.rb
```diff
@@ -101,10 +101,24 @@ def total_number_of_excess_events(excludes = [])
       # Public: Returns number of blackout hours
       #
       # Returns an Integer
-      def number_of_blackout_hours
+      def number_of_power_shortage_hours
         graph.plugin(:merit).order.blackout.number_of_hours
       end
 
+      # Public: Returns the peak of blackout hours, defined as the largest single hour deficit
```
mabijkerk, line 108: Following our discussion that documentation should be useful for both modellers and developers: it would be useful to add units to the comments of these functions: is this MJ, MW, MWh?

louispt1:
@mabijkerk I've addressed your comments by adding the units to the comments in the energy.rb file and shortening the commit hash for the relevant merit commit. We may want to rebase/merge merit before merging this branch and double check the commit hash before merging it here.

app/models/qernel/graph_api/energy.rb
```diff
@@ -93,18 +95,32 @@ def group_of_excess_events(durations = [], excludes = [])
...
-      # Returns an Integer
-      def number_of_blackout_hours
+      # Returns an Integer (hours)
+      def number_of_power_shortage_hours
         graph.plugin(:merit).order.blackout.number_of_hours
       end
 
+      # Public: Returns the peak of blackout hours, defined as the largest single hour deficit
+      #
+      # Returns an Integer (MW)
```
mabijkerk, line 112: @louispt1 I might miss some developer terminology, but integer means it's a whole number right? This returns a number with decimals, shouldn't be a float?

mabijkerk (approved):

louispt1:
These changes have already been merged to master

PR #1568, opened 2025-04-30
Merge household appliances efficiency sliders into one

mabijkerk:
This PR adresses https://github.com/quintel/etmodel/issues/4479. There are currently 8 household appliances efficiency sliders. This seems excessive. They are merged into one.

Should not be merged before the Deploy. The migration in ETEngine still has to be finished. @noracato / @louispt1 it's propsed for next sprint to pick this up. If either of you has time this week, feel free to pick this up. A modeller should review the outcomes.

Goes with: 
- https://github.com/quintel/etsource/pull/3255
- https://github.com/quintel/etmodel/pull/4481

### DIFF
--- a/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
+++ b/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
@@ -21,9 +21,26 @@ class MergeHouseholdAppliancesEfficiency < ActiveRecord::Migration[7.0]
     migrate_scenarios do |scenario|
       next unless SHARE_MAP.values.any? { |key| scenario.user_values.key?(key) }
 
-      # Parent shares are recovered from the csvs in etsource
-      dataset = Atlas::Dataset.find(scenario.area_code)
-      parent_shares = dataset.shares("energy/residences_final_demand_for_appliances_electricity_parent_share")
+      # Skip if the dataset itself can't be found
+      dataset = begin
+        Atlas::Dataset.find(scenario.area_code)
+      rescue Atlas::DocumentNotFoundError, Atlas::DatasetError
+        next
+      end
+
+      # Skip if the share series can't be loaded
+      parent_shares = begin
+        dataset.shares("energy/residences_final_demand_for_appliances_electricity_parent_share")
+      rescue Atlas::ResourceNotFound, Atlas::DatasetError
+        next
+      end
+
+      # Skip if any individual parent‐share key is missing/nil
+      missing = SHARE_MAP.keys.reject do |parent_key|
+        value = parent_shares.get(parent_key) rescue nil
+        !value.nil?
+      end
+      next unless missing.empty?
 
       # Calculate weighted avg. Start value for the old sliders was 0.0. If input is not set, it's treated as 0.0.
       weighted_avg = SHARE_MAP.sum(0.0) do |parent_share_key, input_key|
--- a/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
+++ b/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
@@ -39,7 +39,7 @@ class MergeHouseholdAppliancesEfficiency < ActiveRecord::Migration[7.0]
 
       # The slider values should be checked for their minimum and maximum values
       # TODO: check this! What should be minimum and mamximum?
-      scenario.user_values[NEW_INPUT] = new_input
+      scenario.user_values[NEW_INPUT] = [[new_input, 90.0].min, -90.0].max
     end
   end
 end
--- a/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
+++ b/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
@@ -27,7 +27,7 @@ class MergeHouseholdAppliancesEfficiency < ActiveRecord::Migration[7.0]
 
       # Calculate weighted avg. Start value for the old sliders was 0.0. If input is not set, it's treated as 0.0.
       weighted_avg = SHARE_MAP.sum(0.0) do |parent_share_key, input_key|
-        efficiency = scenario.user_values.fetch(input_key, 0.0)
+        efficiency = scenario.user_values.delete(input_key) || 0.0
         parent_shares.get(parent_share_key) * (1.0 - efficiency / 100.0)
       end
 
--- a/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
+++ b/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
@@ -32,7 +32,7 @@ class MergeHouseholdAppliancesEfficiency < ActiveRecord::Migration[7.0]
       end
 
       # The weighted average is reconverted into a percentage
-      new_input = (weighted_avg - 1) * 100.0
+      new_input = (1 - weighted_avg) * 100.0
 
       # The slider values should be checked for their minimum and maximum values
       scenario.user_values[NEW_INPUT] = [[new_input, 90.0].min, -90.0].max
--- a/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
+++ b/db/migrate/20250429122439_merge_household_appliances_efficiency.rb
@@ -1,0 +1,83 @@
+require 'etengine/scenario_migration'
+
+class MergeHouseholdAppliancesEfficiency < ActiveRecord::Migration[7.0]
+  include ETEngine::ScenarioMigration
+
+  NEW_INPUT = 'households_appliances_electricity_efficiency'.freeze
+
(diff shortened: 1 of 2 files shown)


mabijkerk:
Note: I think we should take into account min and max slider values and incorporate those boundaries in the migration.

noracato:
I'll have time to look at this probably this week, but I do not fully understand the current migration.

A few questions:

1. What was the idea of writing a tempfile during the migration? Why are we dumping all area attributes for every scenario that does not get a new input? Do we need to do something with this data?
2. The parent share: that is the share of a specific appliance node into the households total applicances node. Is this scenario specific or dataset specific? I can see a handful of csvs in etsource with the shares in there, but could it be they have changed in the scenario via an input or something? Or can we just go off the csvs?

mabijkerk:
@noracato I didn't look at the implementation of this migration from @kaskranenburgQ, but my "pseudocode" goal of this migration was the following (see also [this](https://github.com/quintel/etmodel/issues/4479#issuecomment-2837764265) comment).


```ruby

# new input that needs to be set
new_input = households_appliances_electricity_efficiency

# old inputs that need to be merged into a new input, note that these are in percentages
old_inputs = [
    households_appliances_clothes_dryer_electricity_efficiency,
    households_appliances_computer_media_electricity_efficiency,
    households_appliances_dishwasher_electricity_efficiency,
    households_appliances_fridge_freezer_electricity_efficiency,
    households_appliances_other_electricity_efficiency,
    households_appliances_television_electricity_efficiency,
    households_appliances_vacuum_cleaner_electricity_efficiency,
    households_appliances_washing_machine_electricity_efficiency
]

# parent shares that come from the datasets
parent_shares = [
    households_final_demand_for_appliances_electricity-households_appliances_clothes_dryer_electricity@electricity+parent_share,
    households_final_demand_for_appliances_electricity-households_appliances_computer_media_electricity@electricity+parent_share,
    households_final_demand_for_appliances_electricity-households_appliances_dishwasher_electricity@electricity+parent_share,
    households_final_demand_for_appliances_electricity-households_appliances_fridge_freezer_electricity@electricity+parent_share,
    households_final_demand_for_appliances_electricity-households_appliances_other_electricity@electricity+parent_share,
    households_final_demand_for_appliances_electricity-households_appliances_television_electricity@electricity+parent_share,
    households_final_demand_for_appliances_electricity-households_appliances_vacuum_cleaner_electricity@electricity+parent_share,
    households_final_demand_for_appliances_electricity-households_appliances_washing_machine_electricity@electricity+parent_share
]

# the old inputs are converted into factors and multiplied with their respective parent shares to weig them, after which they are summed
weighted_average = [
    households_final_demand_for_appliances_electricity-households_appliances_clothes_dryer_electricity@electricity+parent_share   * (1 - (households_appliances_clothes_dryer_electricity_efficiency / 100.0)),
    households_final_demand_for_appliances_electricity-households_appliances_computer_media_electricity@electricity+parent_share  * (1 - (households_appliances_computer_media_electricity_efficiency / 100.0)),
    households_final_demand_for_appliances_electricity-households_appliances_dishwasher_electricity@electricity+parent_share      * (1 - (households_appliances_dishwasher_electricity_efficiency / 100.0)),
    households_final_demand_for_appliances_electricity-households_appliances_fridge_freezer_electricity@electricity+parent_share  * (1 - (households_appliances_fridge_freezer_electricity_efficiency / 100.0)),
    households_final_demand_for_appliances_electricity-households_appliances_other_electricity@electricity+parent_share           * (1 - (households_appliances_other_electricity_efficiency / 100.0)),
    households_final_demand_for_appliances_electricity-households_appliances_television_electricity@electricity+parent_share      * (1 - (households_appliances_television_electricity_efficiency / 100.0)),
    households_final_demand_for_appliances_electricity-households_appliances_vacuum_cleaner_electricity@electricity+parent_share  * (1 - (households_appliances_vacuum_cleaner_electricity_efficiency / 100.0)),
    households_final_demand_for_appliances_electricity-households_appliances_washing_machine_electricity@electricity+parent_share * (1 - (households_appliances_washing_machine_electricity_efficiency / 100.0))
].sum

# the weighted average is reconverted into a percentage
new_input = (weighted_average - 1) * 100.0
```

noracato:
Thanks! That clears things up!

mabijkerk:
> I see the new min and max will be -90.0 to 90.0: do you want me to round to those if the value exceeds these boundries?

Yes, I added that as one of my earlier comments: we need to check if the new value exceeds the boundaries. If it's higher than the max, it should be set to max. If it's lower than the min, it should be set to the min. Can you add this?

mabijkerk:
@noracato if the share is not set, then its value should be 0. The following code:
```ruby
households_final_demand_for_appliances_electricity-households_appliances_clothes_dryer_electricity@electricity+parent_share   * (1 - (households_appliances_clothes_dryer_electricity_efficiency / 100.0))
```
Then becomes:
```ruby
households_final_demand_for_appliances_electricity-households_appliances_clothes_dryer_electricity@electricity+parent_share   * (1 - (0.0 / 100.0))
```
Which comes down to:
```ruby
households_final_demand_for_appliances_electricity-households_appliances_clothes_dryer_electricity@electricity+parent_share  * 1
```

Is this the way you have included it in your migration?

louispt1:
> @noracato if the share is not set, then its value should be 0. The following code:
> 
> ```ruby
> households_final_demand_for_appliances_electricity-households_appliances_clothes_dryer_electricity@electricity+parent_share   * (1 - (households_appliances_clothes_dryer_electricity_efficiency / 100.0))
> ```
> 
> Then becomes:
> 
> ```ruby
> households_final_demand_for_appliances_electricity-households_appliances_clothes_dryer_electricity@electricity+parent_share   * (1 - (0.0 / 100.0))
> ```
> 
> Which comes down to:
> 
> ```ruby
> households_final_demand_for_appliances_electricity-households_appliances_clothes_dryer_electricity@electricity+parent_share  * 1
> ```
> 
> Is this the way you have included it in your migration?



Testing my understanding as I'm jumping into this issue, but based on how I understand this, the code should become:
```
if scenario.user_values.key?(input_key)
  parent_shares.get(parent_share_key) * (1 - (scenario.user_values[input_key]) / 100.0)
else
  parent_shares.get(parent_share_key)
end
```

Aka, if the input_key is not set, the weighted average will be set to the parent share. 

I've implemented it like that in the latest commit - please let me know if you think I've misunderstood @mabijkerk

louispt1 (approved):
LGTM - we'll have to rebase pre-merging

noracato:
@mabijkerk @louispt1 looks ready to merge!

mabijkerk:
I still want to test the migration on some sample scenarios before merging.

db/migrate/20250429122439_merge_household_appliances_efficiency.rb
```diff
@@ -0,0 +1,41 @@
...
+      parent_shares = dataset.shares("energy/residences_final_demand_for_appliances_electricity_parent_share")
+
+      # Calculate weighted avg. Start value for the old sliders was 0.0. If input is not set, it's treated as 0.0.
+      weighted_avg = SHARE_MAP.sum(0.0) do |parent_share_key, input_key|
+        efficiency = scenario.user_values.fetch(input_key, 0.0)
+        parent_shares.get(parent_share_key) * (1 - efficiency / 100.0)
+      end
+
+      # The weighted average is reconverted into a percentage
+      new_input = (weighted_avg - 1) * 100.0
```
kndehaan, line 35: This is incorrect, it should be:
```
new_input = (1-weighted_avg) * 100.0
```

The new input represents an efficiency improvement, meaning that a positive value results in a decrease in electricity demand. The way it's defined now, causes that a positive value results in increase in electricity demand.

kndehaan:
The migration fails when it encounters a dataset that it cannot find, from which the parent_shares should be retrieved (which are based on the parent dataset):
```
Atlas::DocumentNotFoundError: Could not find a dataset with the key "BU00140201_noorderplantsoenbuurt" (Atlas::DocumentNotFoundError)
```
I assume that this error occurs for all retired datasets that don't live on ETSource anymore. If the dataset cannot be found, it can probably also not find what the parent dataset is. What is the best way to tackle this issue? @mabijkerk @noracato @louispt1 

One solution might be to take the parent shares of nl2019 as default for these cases.

db/migrate/20250429122439_merge_household_appliances_efficiency.rb
louispt1: Nice catch - I've pushed the change

noracato:
One last comment: do we want to remove the user values of the old sliders in a follow up migration?

mabijkerk:
@noracato good catch, can you add that?

PR #1580, opened 2025-05-12
Add migration to remove retired household behaviour keys

mabijkerk:
Household behaviour inputs/sliders were retired. See https://github.com/quintel/etsource/pull/3261. The migration in this PR makes sure that the keys are also removed from all scenarios.

Note: the `households_behavior_close_windows_turn_off_heating` input/slider had already been removed much earlier, during heat modelling changes. The key had never been removed from scenarios though.


noracato (approved):
Beautiful!

PR #1583, opened 2025-05-14
Remove intelligent control input sliders from scenarios

kaskranenburgQ:
This PR solves https://github.com/quintel/etmodel/issues/4490 by removing the intelligent control lighting inputs in the scenarios

Goes together with:
ETSource: https://github.com/quintel/etsource/pull/3264 
ETModel: https://github.com/quintel/etmodel/pull/4491
ETDataset: Removal of dataset keys; TBD

### DIFF
--- a/db/migrate/20250514083809_remove_intelligent_control.rb
+++ b/db/migrate/20250514083809_remove_intelligent_control.rb
@@ -1,0 +1,23 @@
+require 'etengine/scenario_migration'
+
+class RemoveIntelligentControl < ActiveRecord::Migration[7.1]
+  include ETEngine::ScenarioMigration
+
+  # Retired household behaviour keys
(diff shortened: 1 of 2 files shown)


kaskranenburgQ:
This migration causes minor changes in comparison to beta, see the [ii3050v2 NT Leadership scenario](https://beta.my.energytransitionmodel.com/saved_scenarios/5136) in which both removed sliders where set to 95%:
Beta:
Screenshot 2025-05-14 at 12 11 35

Local:
Screenshot 2025-05-14 at 12 12 03

I personally don't see how we could mitigate these changes, since there are no other sliders that influence the useful demand of lighting in the buildings sector. Since these changes are generally small, I feel that no extra effort is required.

Issue #1585, opened 2025-05-19
Number of blackout hours on dashboard should align count of non-zero deficits in data-export


kaskranenburgQ:
The number of blackout hour on the dashboard item can differ from the amount of hours with deficits in the merit order download.

For example, for [this](https://my.energytransitionmodel.com/saved_scenarios/19919) scenario, the front-end shows that there are 98 blackout hours per year. When you download the electricity load curves and count the nonzero deficits in Excel, there are 109 blackout hours.

This issue probably has a similar cause as #1403.

louispt1:
I believe the blackout hours do align in the export csv and in the dashboard item. @noracato you had some concern that that the curtailment and satisfied_demand are excluded in the export, but I think it's an intentional decision to exclude them on the basis that they should not impact the output, as they have no output defined ([see this issue](https://github.com/quintel/etsource/issues/2818)).

Since the hours align, can I close this issue?

PR #1440, opened 2024-06-17
Hybrid Offshore Wind

noracato:
With this PR the modelling of hybrid offshore wind is added to the ETM. This comes with, among others, the following features:

-   Hybrid offshore wind turbine that can generate electricity for HV network or directly for connected offshore electrolyser. 
-   Price-based offshore electrolyser that can receive electricity from hybrid wind turbine or from onshore HV network
-   New adapter classes `HybridOffshore` and `WithSatisfiedDemand` for MeritFacade

Goes with:

https://github.com/quintel/documentation/pull/195
https://github.com/quintel/etsource/pull/3067
https://github.com/quintel/etlocal/pull/524
https://github.com/quintel/etmodel/pull/4290
https://github.com/quintel/etdataset/pull/1001
quintel/merit#154

PR #1589, opened 2025-05-21
Updated load shifting adapter to properly account for deficit hours

### LINKED ISSUE #1535: Price-setting during blackout hours should take DSR deficit limits into account
In the scenario II3050 v2 - International Trade on beta, there are two timesteps in which deficits occur. 
This can be seen in the deficits column in the export "Electricity load curves"
These timesteps are: 

1. 2050-11-15 21:00 --> Deficits of 3519.8903
2. 2050-12-20 06:00 --> Deficits of 627.6659

In the first timestep, the electricity price is 3000.0, this is the "Value of Lost load", and the price that is set when demand can't be matched in the model. 

In the second timestep, the electricity price is 220.8845, these are the marginal costs of the highest producer in the scenario, the node energy_chp_supercritical_ht_waste_mix. 

The difference between the two timesteps seems to lie in the load shifting of Industry in ICT. At timestep 1, there is load shifting at full capacity in ICT, at timestep 2 there is no load shifting in ICT. It is not entirely clear to me why at this timestep there is no load shifting. 

In addition, the blackout hours dashboard item shows there is only one blackout hour. 
I have calculated the deficits manually in the excel by subtracting the total demand from the total production. It produced the same results as the deficits column. 
Since there is a deficit and the node energy_chp_supercritical_ht_waste_mix is price-setting, might there be producers missing from the data export? 

I propose the following steps:

- Investigate why industry ICT does not perform load shifting on the second timestep. 
- Investigate why there is a difference in the blackout hours and the hours with deficits. 

Some background on the dashboard item: 
This is calculated in merit by subtracting the sum of the consumers of the producer curve. It then counts the times that there is a deficit. 
The blackout price is determined in the same way. When there is a deficit, the price will be set to 3000.0. 

It could be that because the industry ICT load shifting is not running, the timestep is not recognized as a blackout-hour, so that the price of the highest running dispatchable is taken as the price. 

Notifying @mabijkerk & @noracato

louispt1:
This refactor enforces a cumulative deficit limit to make sure the cumulative energy deficit is tracked over time. In this way, output should cease once the configured deficit_capacity is reached. The previous implementation treated the load curve as static, ignoring any notion of capacity constraints for shifting demand.

I've also added a small spec for the changes.

@kaskranenburgQ could you test whether these changes have the desired effect?

Closes #1535 (and maybe #1585).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -74,7 +74,7 @@ gem 'parallel'
 gem 'ruby-progressbar'
 
 # own gems
-gem 'quintel_merit', ref: 'd79f11d', github: 'quintel/merit'
+gem 'quintel_merit', ref: '2411622', github: 'quintel/merit'
 gem 'atlas',         ref: '8e854b1', github: 'quintel/atlas'
 gem 'fever',         ref: '2a91194', github: 'quintel/fever'
 gem 'refinery',      ref: 'c39c9b1', github: 'quintel/refinery'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -36,8 +36,8 @@ GIT
 
 GIT
   remote: https://github.com/quintel/merit.git
-  revision: d79f11db131edbb0107542c681781a2a7fae1905
-  ref: d79f11d
+  revision: 2411622e264211d593f2c148ff30484d70cbedf2
+  ref: 2411622
   specs:
     quintel_merit (0.1.0)
       numo-narray


kaskranenburgQ:
I have tested this by using the ii3050v2-NT scenario on beta.
To test this scenario locally I used the beta database. 

On the beta version of this scenario, a deficit occurs on hour 17/01/06:00 (918.9505 MW) with a price of 118.58. 

Running the scenario locally with these changes, the same deficit and same price result came at this hour. 

Let's discuss again @louispt1

louispt1:
Having looked into this issue, I found that the demand/supply within the merit blackout module and the merit_csv_serializer are actually quite different. I don't think the load shifting adapter is at fault here, and I can see that `energy_flexibility_flow_batteries_electricity` discharges quite a lot of energy the very next hour after the deficit occurs (so it has capacity).

[This commit](https://github.com/quintel/etengine/commit/24d79d761296f146374c3e7dc6b839f980a82f3f) as part of [this PR/project](https://github.com/quintel/etengine/pull/1440) excluded satisfied demand from the csv, and on the ii3050v2 National Governance scenario on beta a bunch of nodes are excluded from the csv results, including:

	•	buildings_final_demand_for_space_heating_electricity
	•	energy_battery_solar_pv_solar_radiation
	•	energy_battery_wind_turbine_inland
	•	households_final_demand_for_hot_water_electricity
	•	households_final_demand_for_space_heating_electricity

So the difference between the calculated deficit column in the CSV and the reported power_shortage_hours in the blackout module appears to be by design - do you know why we might exclude some producers in this export @kaskranenburgQ?

At the moment it doesn't seem like any action is actually required, but perhaps we should discuss whether these exclusions are correct and necessary.

PR #1592, opened 2025-06-12
Retire FinalDemand, MaxDemand and DependentSupply recursive methods

louispt1:
This PR retires recursive methods deemed to no longer be useful or sensible given the evolution of the model. I have also removed the relevant spec and retired the 'abroad' and 'domestic_dead_end' functionality as it is no longer relevant. I separated the modules/components into different commits to make the review process more straightforward.

ETSource methods that may no longer be relevant should also be sought out and removed.


noracato (commented):
Looks good to me, but we really need a modeller to check if the outcomes are still valid. Let's ask around who would have time!

PR #1595, opened 2025-06-25
Puma config that draws on the cpu assignment to determine workers

louispt1:
This PR re-draws the puma config in ETEngine to use the cpu assignment of the container to determine workers (if WEB_CONCURRENCY is not set explicitly).

This is more of an 'example' PR for now. Before merging this PR, we should also correctly set the WEB_CONCURRENCY variable or remove it [as per this issue](https://github.com/quintel/ansible/issues/91).

If we would like to move forward with adjusting the puma configurations in our other rails applications, for the thread-safe applications (ETModel and MyETM at least) we can also allow the threads assignment in their puma configurations to be determined buy the cpu_cores assignment, matching the number of threads and workers to the number of cpus available, or explicitly setting the RAILS_MAX_THREADS variable to limit threads explicitly.

```
thread_count = Integer(ENV.fetch('RAILS_MAX_THREADS', cpu_cores.to_s))
threads thread_count, thread_count
```

This PR also introduces `preload_app!` to the puma config which helps save memory between workers using [copy on write](https://en.wikipedia.org/wiki/Copy-on-write). This seems to make sense in any case where more than one worker might be used.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -1,35 +1,31 @@
 # frozen_string_literal: true
 
+require 'etc'
+
 # Puma can serve each request in a thread from an internal thread pool.
 #
 # The `threads` method setting takes two numbers: a minimum and maximum. Any libraries that use
 # thread pools should be configured to match the maximum value specified for Puma. Default is set to
 # 5 threads for minimum and maximum; this matches the default thread size of Active Record.
 #
-# ETEngine is not thread-safe.
+# ETEngine is not thread-safe, so we use 1 thread per worker.
 threads 1, 1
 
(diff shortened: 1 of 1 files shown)


louispt1:
This PR should not be merged. I have set up the correct configurations for the puma config for the rails applications in our environment. PRs for each change can be found here:
- https://github.com/quintel/my-etm/pull/148
- https://github.com/quintel/etengine/pull/1600
- https://github.com/quintel/etmodel/pull/4520
- https://github.com/quintel/projects/pull/301
- https://github.com/quintel/etlocal/pull/604
- https://github.com/quintel/dbbot/pull/7

There are two categories of configuration, one for thread safe and one for non-thread safe applications.

PR #1646, opened 2025-10-24
Add DATASET_CURVE(key) method for GQL lookup

### LINKED ISSUE #1644: Query the air_temperature curve
Currently it's not possible to query some non-uploaded curves, like the `air_temperature`. 
We can add this possibility to `Gql::Runtime::Functions::Curves` or to `Gql::Runtime::Functions::Lookup`. Depending on what type of curves we want to extract either one would make more sense.

Questions:
- Do we want to focus on just `air_temperature` or also on other curves?
- If so, which?

louispt1:
Adding this new GQL method allows curves to be loaded from the dataset by key.

The motivation for this change is the p2h project, where the air_temperature needs to be queried for the purposes of adding the chart in that project.

These changes enable GQL queries like:
```
DATASET_CURVE('weather/air_temperature')
DATASET_CURVE('total_demand')
```

The key can be:
  - A curve set path: 'weather/air_temperature', 'solar/pv'
  - A simple profile name: 'total_demand', 'buildings_heating'
  - A user-attached curve

This PR closes #1644. 

There is an additional consideration - This function relies on the causality curves module, and a graph needs to be established to query the curves this way. In the future, we may consider moving this curves module out of Causality so it can be used more generally like for cases like this. I will open an enhancement issue for this

### DIFF
--- a/app/models/gql/runtime/functions/lookup.rb
+++ b/app/models/gql/runtime/functions/lookup.rb
@@ -327,7 +327,7 @@ module Gql::Runtime
       # Returns an array of 8760 numeric values.
       def DATASET_CURVE(key)
         curves = Qernel::Causality::Curves.new(scope.graph, rotate: 0)
-        curves.send(:load_curve, key, nil).to_a
+        curves.curve(key, nil).to_a
       end
     end
   end


app/models/gql/runtime/functions/lookup.rb
```diff
@@ -312,6 +312,23 @@ def CURVE_SET_VARIANTS(name)
...
+      #
+      # The key can be:
+      #   - A curve set path: 'weather/air_temperature', 'solar/pv'
+      #   - A simple profile name: 'total_demand', 'buildings_heating'
+      #   - A user-attached curve (if configured)
+      #
+      # Returns an array of 8760 numeric values.
+      def DATASET_CURVE(key)
+        curves = Qernel::Causality::Curves.new(scope.graph, rotate: 0)
+        curves.send(:load_curve, key, nil).to_a
```
noracato:
```suggestion
        curves.curve(key).to_a
```
It's a good practice to use public methods, especially when calling from outside the module with user input.

noracato (changes requested):

PR #1647, opened 2025-10-24
Create adapter for temperature based power to heat participants

### LINKED ISSUE #1632: Add behaviour for temperature setpoint and availability curves
We need to create a new adapter in the Merit Facade. The current p2h connected to the heat network act as merit flex generics, but we need to add some extra behaviours.

## Atlas

In Atlas we have to allow a new merit attribute to be set. This attribute, when set on a node, can be changed by a new input, and will be picked up by the new adapter.

```
- merit_order.temperature_cutoff = 15.0
```

This is a simple solution, that **will** make it necessary to update that attribute on each affected node when creating an input.

## TemperatureBasedPowerToHeatAdapter

The adapter will inherit from `FlexAdapter::Base` with the following extensions:

```ruby
# Internal: The curve of air temperatures in the region.
def temperature_curve
  @context.curves.curve('weather/air_temperature', @node)
end
```

Then we need to create an availability curve based on the temperature cutoff, **OR** we use a user uploaded curve:

```ruby
def availability_curve
  user_uploaded_curve || calculated_curve(@node.config.temperature_cutoff)
end
```

We can make use of existing classes in Merit, so no need for any changes in the gem

```
Merit::Flex::VariableConsumer
# A flexibility participant which can only consume, and whose input capacity may vary each hour
# using an availability curve. This differs from other flexibility options where input capacity
# and availability are fixed for the whole year.
```

## ETSource
Add the new availability curve to be allowed to be uploaded, and which inputs it will cancel.
In user_curves config make sure the curve is uploadable without being in the dataset, disable the temperature setpoint input

## Notes

The availabilty curve is not downloadable/queryable

We might solve the weather behaviour issue (effect of raising global temp on standard weather curve) in a separate issue

noracato:
Implements behaviour specific to the power2heat (pumps) that are dependent on outside temperature.

The participant will be fully available when the outside temperature is above the temperature_cutoff point (in degrees Celsius). And will have an availability of 0, when below this point.

Users can also set their own availabilty curves on the node to overwrite this behaviour.

Uses `Merit::Flex::VariableComsumer` as this class already consumes based on an availability curve.

Closes #1632

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -75,7 +75,7 @@ gem 'ruby-progressbar'
 
 # own gems
 gem 'quintel_merit', ref: '54d2be1', github: 'quintel/merit'
-gem 'atlas',         ref: 'e914ca0', github: 'quintel/atlas'
+gem 'atlas',         ref: '34f6d2b', github: 'quintel/atlas'
 gem 'fever',         ref: '2a91194', github: 'quintel/fever'
 gem 'refinery',      ref: 'c39c9b1', github: 'quintel/refinery'
 gem 'rubel',         ref: 'e36554a', github: 'quintel/rubel'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GIT
   remote: https://github.com/quintel/atlas.git
-  revision: e914ca0f5b5a9f4e1e46a4803a82751a1daa2f81
-  ref: e914ca0
+  revision: 34f6d2b877b91edf636f449eb1396f6329d63614
+  ref: 34f6d2b
   specs:
     atlas (1.0.0)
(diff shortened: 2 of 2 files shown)


app/models/qernel/merit_facade/temperature_based_power_to_heat_adapter.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+module Qernel
+  module MeritFacade
+    # Implements behaviour specific to the power2heat (pumps) that
```
louispt1:
```suggestion
    # Implements behaviour specific to the power2heat (pumps & boilers) that
```

PR #1654, opened 2025-10-29
Fix ScenarioUser coupling bug

louispt1:
This PR:
- Fixed bug in ScenarioUser.couple_existing_user which was looking up by user_id rather than user_email
- Expanded test coverage: added spec covering the user coupling logic more fulsomely

## Motivation:
Users added as collaborators via MyETM were not being properly linked to their accounts in ETEngine. The couple_existing_user hook never actually coupled users because it looked up User.find_by(id: user_id)
  when user_id was nil. This meant:
  - Existing users added by email weren't immediately coupled to their account
  - Collaboration permissions weren't working across MyETM and ETEngine

---

### Testing
Added spec to check:
  - Existing users are automatically coupled when added by email
  - Non-existent users remain pending with user_email set
  - Duplicate user detection works correctly via unique indices
  - User signup triggers automatic coupling of pending ScenarioUser records
  
---

### Open questions:
- Do we want to show pending users (as in, scenario user with email but no id) on the etengine inspect scenario view?

Goes with this [MyETM PR](https://github.com/quintel/my-etm/pull/177)

### DIFF
--- a/app/models/scenario_user.rb
+++ b/app/models/scenario_user.rb
@@ -16,7 +16,7 @@ class ScenarioUser < ApplicationRecord
   def couple_existing_user
     return unless user_email.present? && user_id.blank?
 
-    couple_to(User.find_by(id: user_id))
+    couple_to(User.find_by(user_email: user_email))
   end
 
   # Couples the record to an existing User.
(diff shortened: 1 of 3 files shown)


noracato (approved):
Nice catch!

PR #1657, opened 2025-11-03
Scenario updater

### LINKED ISSUE #1186: Validate inputs when a scenario is created
I believe we only validate inputs when updating a scenario and that *new* scenarios are not subject to any checks on the user values.

Validation of inputs means:

* Asserting that the inputs given really exist.
* Asserting that the given value is valid (either within the min/max for normal inputs, or is a permitted value for enum inputs).
* Asserting that share groups sum to 100%.

This validation not being done makes some sense from the perspective of the ETM:

* New scenarios created in the ETM never have any inputs.
* Scenarios created with a preset will have correct share groups, as the preset will have been validated.

It's not very nice for API users though, who may choose to set inputs when they create a scenario. Currently, this validation lives in [`ScenarioUpdater`](https://github.com/quintel/etengine/blob/9a4deb26af4dfbecc8efae51b74bd6678ae03b66/app/models/api/v3/scenario_updater.rb).

I'm reasonably certain `ScenarioUpdater` would work with an unsaved scenario. It would also be terrific to split this class into smaller classes, each with fewer responsibilities; it's currently rather large and doing too many things.

### LINKED ISSUE #1602: Setting external coupling inputs via the engine doesn't work
For a blank nl2019 I try to set an external coupling input via the engine, but get the following error:

Image

When I try to set a 'new' external coupling input through the front-end in the II3050v2 National Leadership 2050 scenario (which is already coupeled), this works. However, when I try to set the same external coupling input via the engine for this scenario, I get the same error as above.

There are no issues with setting the external coupling input via the stable #2025-01 engine. Note in the screenshot above that it formats the user values and balanced values differently when the page with error message is returned. Considering that everything via the front-end works fine and seeing the strange format that is applied to user and balanced values, might this issue have something to do with the new data dump download functionality (which is not available on the stable #2025-01 engine)?

@noracato @louispt1

### LINKED ISSUE #1608: Align updating inputs via the API and interface
When users update the user values of a scenario via the interface, we do not use the same instruments as via the API. The inputs are barely validated, and just get updated as is. 

This has some advantages, like the possibility to remove outdated inputs. But also allows interface users to accidentally break scenarios with malformed inputs, and gives strange errors from time to time. Like #1602. 

We should start to use some method from the API route to align the interface and API more.

### LINKED ISSUE #1626: Input validation - reevaluation
This overarching issue will collect all issues concerning `ScenarioUpdater` and whether or not it's called using different API's and interfaces.

### LINKED ISSUE #1638: Set up new ScenarioUpdater structure
Based on our Miro brainstorming and refining the issue further, I came up with the following file structure:

```
├── base.rb
├── attributes.rb
├── preset_handler.rb
├── orchestrator.rb
└── inputs/
    ├── couplings_manager.rb
    ├── update.rb
    ├── validator.rb
    ├── balancer.rb
    └── balance_validator.rb
```

- [x] Create (models/scenario_updater/base.rb)
```
module ScenarioUpdater
  class BaseComponent
    
    attr_reader :scenario, :params, :current_user
    
    def initialize(scenario, params, current_user)
      @scenario = scenario
      @params = params
      @current_user = current_user
    end
  end
end
```
Consider: What else is common to the base component? Perhaps a definition of TRUTHY values.

- [x] Extract Attributes Component to models/scenario_updater/attributes.rb
```
module ScenarioUpdater
  class Attributes < BaseComponent
    
    def attributes_to_apply
      scenario_data = (params[:scenario] || {}).with_indifferent_access
      
      scenario_data
        .except(:area_code, :end_year, :set_preset_roles, :user_values)
        .merge(metadata: metadata)
    end
    
    private
    
    def metadata
      scenario_data = (params[:scenario] || {}).with_indifferent_access
      
      if scenario_data.key?(:metadata)
        scenario_data[:metadata]
      else
        scenario.metadata.dup
      end
    end
  end
end
```
EDIT: Removed the metadata size validation from the updater as it's already in the model - the model should validate and tell us if the metadata is invalid in which case the scenario update transaction will fail and roll back

- [x] Test the Attributes component
There is a lot of spec that can be re-organised / repurposed.

Steps implemented in [this commit](https://github.com/quintel/etengine/commit/5e3ab15afa5df9bf7cc22e1a8051772d2faf5ae5)

### LINKED ISSUE #1639: (ScenarioUpdater::Inputs) Separation of Scenario Updater Input Validation and Balancing
There should be a dedicated submodule within the ScenarioUpdater for handling inputs:

```
└── inputs/
    ├── couplings_manager.rb
    ├── update.rb
    ├── validator.rb
    ├── balancer.rb
    └── balance_validator.rb
```

1. Extract Input validation logic
Validate:
- [x] Existence
- [x] Numeric
- [x] Boolean
- [x] Enum
- [x] Min/max
- [x] Step values (see [this issue for reference](https://github.com/quintel/etengine/issues/1606))
- [x] Write spec for the Validator that implements these checks

Further discussion is needed on how to handle the step value validation - do we want to coerce values which are in between steps or would we rather throw an error if someone tries to update a value to something in between the step values? At the moment coercion is implemented in [this commit](https://github.com/quintel/etengine/commit/b832e16fa34bb20d9918866d3ff99cf1298aa78c) that implements the Inputs Validator.

2. Inputs Balancer
This will be an interface with the Balancer model:
- [x] calculate_balanced_values
- [x] autobalance
- [x] force_balance

3. Inputs Balance Validation
- [x] validate groups balance

4. Couplings Manager
- [x] activate_from_provided_values
- [x] uncoupled_inputs
- [x] uncouple? (private)

Steps 2-4 and tests have been implemented in [this commit](https://github.com/quintel/etengine/commit/10533cd7997be8a09c061b208d0fa694cfa92c9f)

5. Inputs coordination: ScenarioUpdater::Inputs::Update
This coordination module should process user values
- [x] calculate user values (reset if necessary, handle base user values and parents)
- [x] calculate balanced values (defer to the balancer)
- [x] validate everything (defer to input validation and balance validation from this submodule)
- [x] Integration Tests

See [this commit](https://github.com/quintel/etengine/commit/35fbd62fd128dc49434c97cebb116011ed65567e) for the update class and tests.

6. Write/ repurpose tests 
- [x] Each component should be tested individually based on its single responsibility. Inputs::Update should handle integration testing.

7. Documentation

See [this commit](https://github.com/quintel/etengine/commit/8a1cab92b96057c28228fcc8966a45ad75c13fb8) for addition of docstrings minimally describing each class and the spec.

Note: We should retire the original scenario updater tests and the tests that have been made redundant by the Scenario Updater refactor once the Scenario Updater is fully integrated.

Open TODOs:
- [ ] Decide how to handle step values, at the moment the coercion approach has been implemented, but this is still very much open for discussion.
- [ ] Retire original tests/methods once full implementation has been completed. 

Related to #1606

### LINKED ISSUE #1640: Extract preset handling logic from current ScenarioUpdater
Extract preset handling logic from current ScenarioUpdater and fit into new approach.

Goal: Set up ScenarioUpdater::Preset and tests

Todo:
- [ ] Determine if roles should be copied (set_preset_roles)
- [ ] Copy roles method
- [ ] Tests

### LINKED ISSUE #1641: Use new ScenarioUpdater for create and update
The first step to utilise the ScenarioUpdater changes is to make use of the new ScenarioUpdater `orchestrator` to perform the update and create actions on the API `scenarios_controller`.

1. Refactor the `update` method on the controller
- [ ] Refactor the method to use the `orchestrator`
- [ ] Make any adjustments necessary to the `ScenarioUpdatesSerializer`
- [ ] Test with current setup to see if the changes are breaking (which they should not be)
- [ ] Check the current `update` tests to see if they need to be updated.

2. Refactor the `create` method on the controller
- [ ] Refactor the method to use the `orchestrator` to validate inputs on create (see: #1186) 
- [ ] Refactor the method to use the `orchestrator` to  activate couplings
- [ ] Test with current setup to see if the changes are breaking (which they should not be)
- [ ] Check the current `create` tests to see if they need to be updated.

### LINKED ISSUE #1642: Use new ScenarioUpdater in the Inspect Controller
This issue needs further refinement.

The main idea is to consider how we ought to validate updates made via the inspect controller. 

We first need to decide whether these updates should be validated in the same way and with the same scrutiny as updates via the API - maybe we want to consider ignoring minimum / maximum values or step values when updating inputs via the Inspect Controller? I would argue these updates should be validated just the same as any other for consistency.

The next step is to use the new ScenarioUpdater to handle updates coming from the Inspect controller. This may include:
- [ ] Coercion of data structures
- [ ] Helpful feedback when an update is not permitted

One thing to consider is that by tightening these validation rules for updating scenarios, many scenarios in the Inspect controller will already be invalid when a user goes to edit it. That could be a frustrating experience - for example if you edit one user value, and the scenario throws 10 errors on other values, or coerces a bunch of values based on their step values and the results change. We should consider how to mitigate this issue.

louispt1:
## Description

This PR refactors the existing scenario updater into its own model with sub-services. The new set up is a functional update pipeline that can be used for both the create and update actions with respect to user inputs, couplings, balanced values and parameters for scenarios. 

## Issues Addressed
These changes address a lot of issues relating to scenario validation and aligning the API and inspect controllers. Because the changes are quite extensive, I will explain them with reference to the parts of the code that changed and the issues addressed - as well as explaining some of the design choices. At first I just refactored the existing update logic into a DRY/OOP version of the updater, but after discussion with @noracato we decided using a monadic process with services would make more sense.

The PR: 
- closes #1608 
- closes #1638 
- closes #1639 
- closes #1640 
- closes #1641 
- closes #1642
- closes #1186
- closes #1602

These issues relate to refactoring the 'old' scenario updater by separating concerns and splitting the handling logic into logical compartments. They also relate to removing the old scenario updater and updating the api scenarios controller and inspect scenarios controller to use the new scenario updater setup. Many of the issues are all part of [#1608](https://github.com/quintel/etengine/issues/1608), which is about aligning the update method between the inspect and api controllers.

Settings external couplings as per #1602 now works again.

This is the meta issue about restructuring the scenario updater. It can be considered closed if all other issues tagged are effectively completed.
Closes #1626

## Other changes
Aside from addressing a number of issues, during development further issues were identified and addressed as part of this PR.

For instance, when validating inputs via the inspect controller, all the inputs are re-validated, because instead of settings one or two, you effectively reset all of the values to whatever is in the yaml. That means for existing scenarios which have gone through many iterations of being updated without validation (i.e. the featured scenarios), there are a lot of validity errors: groups don't balance, inputs don't exist etc etc.
To facilitate 'admin usage' in the ETEngine backend, and avoid everyone having to update all the issues in the featured scenarios, we decided to list the errors, but provide users with the option to 'update anyway':

image

image

Another issue I found was that when you start to validate all the inputs being updated via the inspect controller, it becomes very time consuming and expensive to fetch these inputs from the cache for validation. The Input.cache had not been updated in some time, and relied on fetching all inputs. I updated the Input.cache to include the read_many() method which would fetch multiple cache keys at once, and only calculate values for the specific inputs needed. This led to massive speed gains for the inspect scenarios controller in validating inputs.

Finally, the refactor changed the usage of ScenarioUpdater from the older object setup to a monadic setup with service objects. This makes the success/failure paths more explicit, and allows the process to be split into logical 'phases', which also means if validation fails, the pipeline will 'short-circuit' before moving forward with the update phase (for example). This way validation, computation and mutation are separated logically.

### Other comments
Although there are many specs, most of which are taken from the old scenario updater spec to the new scenario updater spec, the services themselves are still only tested quite minimally. Further testing could definitely be beneficial for the ScenarioUpdater services.

@noracato once you're generally happy with the implementation, I think we will need to include the modellers for some thorough testing on their behalf, and to consider the implications of some of these changes.

### DIFF
--- a/app/controllers/inspect/scenarios_controller.rb
+++ b/app/controllers/inspect/scenarios_controller.rb
@@ -166,14 +166,25 @@ module Inspect
       force_update = params[:force_update].present?
       result = ::ScenarioUpdater.new(scenario, updater_params, current_user, skip_validation: force_update).call
       success = false
+      sortable_attrs = user_sortable_attributes
+      sortable_records = build_sortables_hash(scenario)
 
       Scenario.transaction do
+        prepared_sortables, sortable_errors = prepare_sortables(sortable_records, sortable_attrs, force_update: force_update)
+        formatted_sortable_errors = format_sortable_errors(sortable_errors)
+
         if result.failure?
-          handle_updater_failure(result.failure, force_update)
+          combined_errors = Array(result.failure) + formatted_sortable_errors
+          handle_updater_failure(combined_errors, force_update)
           raise ActiveRecord::Rollback
         end
 
-        persist_sortables!(user_sortable_attributes, build_sortables_hash(scenario))
+        if sortable_errors.any?
+          handle_updater_failure(formatted_sortable_errors, force_update)
+          raise ActiveRecord::Rollback
+        end
+
+        persist_sortables!(prepared_sortables)
         success = true
       end
 
@@ -210,20 +221,36 @@ module Inspect
       sortables
     end
 
-    # Persists user sortables with new order values
-    def persist_sortables!(attrs, records)
-      records = records.select { |key, _| attrs.key?(key) }
+    # Applies user input to sortable records, returning the mutated records and any validation errors
+    def prepare_sortables(records, attrs, force_update:)
+      selected_records = {}
+      errors = {}
 
       records.each do |key, record|
+        attributes = attrs[key] || attrs[key.to_s]
+        next unless attributes
+
+        selected_records[key] = record
+
         # Assign the sortable to the scenario explicitly, so we can preserve the object (and errors)
         record.scenario.public_send("#{key}=", record) if record.scenario.respond_to?("#{key}=")
-        record.order = attrs[key][:order].to_s.split
+        record.order = attributes[:order].to_s.split
+
+        next if record.valid?
+
+        if force_update
+          record.order = record.useable_order
+          record.errors.clear
+        else
+          errors[key] = record.errors.full_messages
+        end
       end
 
-      # Validate all records and raise if any are invalid
-      raise ActiveRecord::RecordInvalid unless records.all? { |_, rec| rec.valid? }
+      [selected_records, errors]
+    end
 
-      # Persist or destroy records depending on whether they're default
+    # Persists user sortables with new order values
+    def persist_sortables!(records)
       records.each do |_, record|
         if record.default?
           record.destroy unless record.new_record?
@@ -233,10 +260,29 @@ module Inspect
       end
     end
 
+    SORTABLE_LABELS = {
+      forecast_storage_order: 'Forecast storage order',
+      hydrogen_supply_order: 'Hydrogen producer order',
+      hydrogen_demand_order: 'Hydrogen flex demand order',
+      heat_network_order_ht: 'Heat network (HT) dispatchables order',
+      heat_network_order_mt: 'Heat network (MT) dispatchables order',
+      heat_network_order_lt: 'Heat network (LT) dispatchables order',
+      households_space_heating_producer_order: 'Households space heating producer order'
+    }.freeze
+    private_constant :SORTABLE_LABELS
+
+    def format_sortable_errors(errors)
+      errors.flat_map do |key, messages|
+        label = SORTABLE_LABELS[key] || key.to_s.humanize
+        Array(messages).compact.map { |message| "#{label}: #{message}" }
+      end
+    end
+
     # Handles updater validation failures by setting flash messages and preserving form input
     def handle_updater_failure(errors, force_update)
+      error_array = Array(errors).flatten
       # Store errors for display in the view
-      flash.now[:validation_errors] = errors
+      flash.now[:validation_errors] = error_array
 
       # Preserve raw form inputs for re-rendering (prevents YAML->JSON conversion)
       scenario_params = params[:scenario] || {}
@@ -246,10 +292,10 @@ module Inspect
 
       # If force_update is not set, give user option to force the update
       if !force_update
-        flash.now[:show_force_update] = true
-      else
+        flash.now[:show_force_update] = error_array.present?
+      elsif error_array.present?
         # If force_update was set but still failed, show error without force option
-        error_list = Array(errors).map { |msg| "• #{msg}" }.join("<br>")
+        error_list = error_array.map { |msg| "• #{msg}" }.join("<br>")
         flash.now[:alert] = "Failed to update:<br>#{error_list}".html_safe
       end
     end
--- a/app/views/inspect/scenarios/_form.html.haml
+++ b/app/views/inspect/scenarios/_form.html.haml
@@ -1,16 +1,17 @@
-- if flash[:show_force_update]
+- if flash[:validation_errors].present?
   .alert.alert-warning
-    %strong Validation errors occurred:
+    %strong= flash[:show_force_update] ? 'Validation errors occurred:' : 'Sorry, there were validation errors:'
     %ul
       - Array(flash[:validation_errors]).each do |error|
         %li= error
-    %p
-      You can choose to:
-      %ul
-        %li
-          %strong Edit the values above to fix the errors, then submit normally
-        %li
-          %strong Force update anyway (click the "Force Update" button below to bypass validation)
+    - if flash[:show_force_update]
+      %p
+        You can choose to:
+        %ul
+          %li
+            %strong Edit the values above to fix the errors, then submit normally
+          %li
+            %strong Force update anyway (click the "Force Update" button below to bypass validation)
 - elsif @scenario.errors.any?
   .alert.alert-danger
     %strong Sorry, there were errors updating the scenario:


app/models/scenario_updater/share_groups.rb
```diff
@@ -0,0 +1,16 @@
...
+  class ShareGroups
+    def self.each(values)
+      group_names = values.keys.filter_map do |key|
+        Input.get(key)&.share_group&.presence
+      end.uniq
+
+      Input.inputs_grouped.slice(*group_names).each do |name, inputs|
+        yield name, inputs
+      end
+    end
```
noracato: I feel like `Input.inputs_grouped` already does this:

```ruby
# input.rb l:53-55

def self.inputs_grouped
  @inputs_grouped ||= Input.with_share_group.group_by(&:share_group)
end
```

noracato (commented):
Looks good!

app/models/scenario_updater/share_groups.rb
louispt1: I actually had another look and I think I still like the current implementation. 

The method as it stands calls Input.get to resolve each incoming key and then uses Input.inputs_grouped.slice(*group_names) so only the share groups present in the incoming payload are processed.

Switching to something like Input.inputs_grouped.each would make the updater touch every defined share group, which is not the intended behaviour.

Do you see what I mean @noracato ? Or did you mean something different

kndehaan:
Final remark: an enhancement would be to not permit users to set the `..._user_curve` inputs by hand (for example `flh_of_energy_power_wind_turbine_inland_user_curve`), via both the engine UI and API. They should only be allowed to be set by uploading a custom profile. 

Tested the API as well, nothing strange happening there so that's looking good!

PR #1660, opened 2025-11-05
Guard atlas cache cleanup against missing files

### LINKED ISSUE #1658: Staring a ETEngine server locally fails only the first time.
## Bug Description

> Recently found out other users are having the same issue when booting up ETEngine locally, the first time fails consistently, but it works when booting it up again.

Image

---

## Expected interaction
What is the expected behaviour?

> It should work well the first time around.

---

### Proposed solutions

> Seems like it's looking for a temporary file/folder related to atlas but it doesn't find it, see the errors text below:
```
14:40:46 css.1  | bin/rails aborted!
14:40:46 css.1  | Errno::ENOENT: No such file or directory @ apply2files - /Users/andreu/Repos/etengine/tmp/atlas/lazy/ES25_hart_van_brabant.pack (Errno::ENOENT)
```
```
14:40:46 css.1  | Caused by:
14:40:46 css.1  | Errno::ENOTDIR: Not a directory @ dir_s_rmdir - /Users/andreu/Repos/etengine/tmp/atlas/lazy/ES25_hart_van_brabant.pack (Errno::ENOTDIR)
```

louispt1:
## Description
Previously, the expire_all! method would fail silently if file deletions raised exceptions (e.g., missing files or permission errors). The updated version ensures failures are logged without interrupting the cleanup process.

When the app is started in dev mode, both the web and css/tailwinds processes load the rails environment and call etsource.rb, so Etsource::Dataset::Import.loader.expire_all! is called by both processes. Potentially there is a race condition causing issues with refreshing the cache that the begin/rescue block avoids.

For me at least this solved the problem - would be interested to see if that works in general / for you @aaccensi 

Closes #1658

### DIFF
--- a/app/models/etsource/atlas_loader.rb
+++ b/app/models/etsource/atlas_loader.rb
@@ -64,7 +64,15 @@ module Etsource
       #
       # Returns nothing.
       def expire_all!
-        Pathname.glob(@directory.join('*.pack')).each(&:delete) if @directory.directory?
+        return unless @directory.directory?
+
+        Pathname.glob(@directory.join('*.pack')).each do |path|
+          begin
+            path.delete
+          rescue => e
+            warn "Failed to delete #{path}: #{e.message}"
+          end
+        end
       end
 
       private


noracato (commented):
Did you check if this could affect production?

louispt1:
> Did you check if this could affect production?

That's a good point - I suppose the risk would be that real errors are rescued on production, and stale .pack files could be left on disc. This issue also shouldn't occur on pro in the first place, so I actually think the change doesn't make sense.

Maybe a simpler solution is to just delete the tmp/atlas/lazy files you have locally and then I don't think the issue should reoccur.

This ties in to your comment @aaccensi:
> I do not know how to consistently reproduce the error

I think once you've cleared the cache once, this issue is resolved, which makes it a little hard to replicate. Let me know what you think @noracato @aaccensi - I cannot trigger the error again anymore and I doubt it will reoccur unless we delete datasets again which we do quite rarely

louispt1:
Closing - I don't think this is a good idea

PR #1670, opened 2025-11-19
Api collections

### LINKED ISSUE #1664: API Create/Update collection routes are not working
## Bug Description

The create and update collection routes through the ETEngine API seem to not work. The other routes, index, show and even destroy work.

---

### Steps to Reproduce

1. Run MyETM and ETEngine locally.
2. Make a create or update call according to the API documentation, like the
     following curl (just make sure the scenario_ids are correct in your case):
 ```
 curl -X POST "http://localhost:3000/api/v3/collections" \
  -H "Accept: application/json" \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $ETM_API_TOKEN" \
  -d '{"title":"New collection name", "scenario_ids":[47]}'
 ```
 4. We just get the response:`{"errors":["Not found"]}`
---

## Expected interaction

Those routes are a kind of pass-thru so they should work very similarly to their equivalent routes in the MyETM API.

---

### Proposed solutions

In my investigation the issue seems to originate in the authorization process for update.
create routes in Api::V3:CollectionController:
```
      before_action(only: %i[create update]) do
        authorize!(:write, Scenario)
      end
```

aaccensi:
## Description

The create and update collections via the API has been fixed by bringing it closer in concept to a simple pass-thru towards MyETM. Yet a couple of actions are still happening on ETEngine to make the API more friendly:
- We do not need to use the `version` parameter, it can be used but otherwise it is simply defaulted from the settings.
- In the very same manner, the `interpolation` parameter can be used but otherwise it is simply defaulted to false.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Enhancement
- [ ] Documentation

## Checklist

## Related Issues

Closes #1664

### DIFF
--- a/app/controllers/api/v3/collections_controller.rb
+++ b/app/controllers/api/v3/collections_controller.rb
@@ -28,7 +28,7 @@ module Api
       end
 
       def show
-        response = my_etm_client.get("/api/v1/collections/#{params.require(:id).to_i}")
+        response = my_etm_client.get("/api/v1/collections/#{params.require(:id)}")
         render json: response.body
       rescue Faraday::ResourceNotFound
         render_not_found
@@ -39,7 +39,8 @@ module Api
           endpoint_path: '/api/v1/collections',
           method: :post
         ).call(
-          params: collection_params({ version: Settings.version_tag, interpolation: false }),
+          params: collection_params({ interpolation: false }),
+          ability: current_ability,
           client: my_etm_client
         ).either(
           ->((data, *)) { render json: data },
@@ -52,7 +53,8 @@ module Api
           endpoint_path: "/api/v1/collections/#{params.require(:id).to_i}",
           method: :put
         ).call(
-          params: collection_params(),
+          params: collection_params,
+          ability: current_ability,
           client: my_etm_client
         ).either(
           ->((data, *)) { render json: data },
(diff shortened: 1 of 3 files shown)


app/controllers/api/v3/collections_controller.rb
```diff
@@ -39,8 +39,7 @@ def create
           endpoint_path: '/api/v1/collections',
           method: :post
         ).call(
-          params: params.permit!.to_h.merge(version: Settings.version_tag),
-          ability: current_ability,
+          params: collection_params({ version: Settings.version_tag, interpolation: false }),
```
noracato: Double check `interpolation: false`

app/services/upsert_transition_path.rb
```diff
@@ -21,51 +21,19 @@ def initialize(endpoint_path:, method:)
...
-    accessible_ids = scenarios.map(&:id)
-
-    if accessible_ids == ids.sort
-      Success(accessible_ids)
-    else
-      missing_ids = ids - accessible_ids
-      Failure(scenario_ids: missing_ids.index_with { |_id| ['does not exist'] })
-    end
-  end
-
```
noracato: Keep for backwards compatibility

app/services/upsert_transition_path.rb
```diff
@@ -21,51 +21,19 @@ def initialize(endpoint_path:, method:)
...
   end
-
-  # # Decorates the params with the scenario's area code and end year when a scenario_id is set.
-  def full_params(params, scenarios)
-    return params if scenarios.blank?
-
-    # The highest end year is always the last scenario.
-    latest_scenario = scenarios.last
-    params.merge(area_code: latest_scenario.area_code, end_year: latest_scenario.end_year)
-  end
```
noracato: Yes, but merge with `version` instead?

PR #1671, opened 2025-11-20
Account for treat as loss edges in output compensation 

### LINKED ISSUE #1667: Edges with `treat_as_loss` set to true, shoudl be treated as loss by `RecursiveFactor`
`RecursiveFactor` uses the following method to account for losses and efficiencies when calculating for instance primary demand:

```ruby
def output_compensation_factor
  loss_compensation_factor * output_efficiency_compensation_factor
end
```

We should extend the `loss_compensation_factor` with a new factor `treat_as_loss_compensation_factor`.

Treat-as-loss edges should be accounted for based on
- the `slot` share of their carrier
- the share of the `edge` in the carriers `slot`

louispt1:
## Description
These changes introduce `treat_as_loss` as an optional attribute on edges. A node method: `treat_as_loss_output_conversion` has been added to calculate the portion of output going to edges marked as treat-as-loss by combining the carrier share of total output (slot conversion) and the edge share within that slot. Finally, that factor is used in the recursive calculations here: `output_compensation_factor` to determine how losses need to be accounted for given the treat_as_loss edges are excluded from the calculation.

## Outcome
  1. Edges marked `treat_as_loss = true` are excluded from recursion (unused heat node gets 0 primary demand)
  2. Siblings are compensated via `treat_as_loss_compensation_factor` (consumer gets 100% primary demand)

Closes #1667

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -75,7 +75,7 @@ gem 'ruby-progressbar'
 
 # own gems
 gem 'quintel_merit', ref: '54d2be1', github: 'quintel/merit'
-gem 'atlas',         ref: '603c2dd', github: 'quintel/atlas'
+gem 'atlas',         ref: '8db16bb', github: 'quintel/atlas'
 gem 'fever',         ref: '2a91194', github: 'quintel/fever'
 gem 'refinery',      ref: 'c39c9b1', github: 'quintel/refinery'
 gem 'rubel',         ref: 'e36554a', github: 'quintel/rubel'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GIT
   remote: https://github.com/quintel/atlas.git
-  revision: 603c2dd4cf704b40eb6bd0e99c660e902b4ea7b3
-  ref: 603c2dd
+  revision: 8db16bb4022d568002073631b418458749bb441d
+  ref: 8db16bb
   specs:
     atlas (1.0.0)
       activemodel (>= 7)


mabijkerk:
Thanks for checking this @noracato.

- primary_co2_emission_of_bio_carriers: good, this should be treated the same as normal emissions.
- demand_of_bio_resources_including_abroad_of_carrier: good, this should work similar to primary demand
- dependent_supply_of_carriers: we do not use this anymore, not relevant if it changes
- final_demand: we do not use this anymore, not relevant if it changes
- final_demand_of_carrier: we do not use this anymore, not relevant if it changes
- primary_demand_co2_per_mj_of_carrier: good, this should be treated the same as normal emissions.

kndehaan:
One remark: final_demand_of_carrier is still used in queries for the charts [use_of_final_demand_in_industry_aluminium](https://github.com/quintel/etmodel/blob/master/config/interface/output_element_series/use_of_final_demand_in_industry_aluminium.yml) and [use_of_final_demand_in_industry_other_metals](https://github.com/quintel/etmodel/blob/master/config/interface/output_element_series/use_of_final_demand_in_industry_other_metals.yml). 

When I look at [final_demand.rb](https://github.com/quintel/etengine/blob/e2c921c327389f8f8c3d5dd470ae33ba41c69006/app/models/qernel/recursive_factor/final_demand.rb), it interpret that it traverses back until it encounters the `final_demand_group` on an edge and takes its value. Our changes in the recursive method will not affect the final_demand_of_carrier method. 

I tested this locally: the results of the queries are not affected by the `treat_as_loss` attribute. 

It might, however, be good to update the queries of above mentioned charts to entirely phase out the use of the final_demand_of_carrier method. I opened an issue for this: https://github.com/quintel/etsource/issues/3361

PR #1672, opened 2025-11-28
Allow accesstokens from query string from etmodel passthru

noracato:
ETModels passthru was sending access tokens in the query string as they can't be set in proper headers. 
But the engine was not picking these up. This caused that result data csvs could not be downloaded for private scenarios in etmodel.

I added an extra clause that will check for a token in the query string.

Notifying @DorinevanderVlies


louispt1 (approved):
LGTM

PR #1683, opened 2025-12-12
Handle setting preset scenario roles on create correctly

louispt1:
## Description
This PR supports [the changes in pyetm to be able to set preset roles on create](https://github.com/quintel/pyetm/pull/108).

The set_preset_roles param needs to not be set directly on the Scenario model, but rather handled explicitly by the ScenarioUpdater.


louispt1:
> Was this already broken? Also from etmodel and my-etm calls?

I think so - it's hard because this changed to be a lot more explicit in the new Scenario Updater approach, so it could have broken in the update, or it could have already been broken and now we can just tell it is

PR #1685, opened 2025-12-17
Interpolate between two scenarios

### LINKED ISSUE #1684: Interpolate between two scenarios
@aaccensi I had a go at refining the changes required for interpolation:

## Description
Refactor the interpolation setup to support interpolating between two scenarios (a start scenario and an end scenario) rather than from default values to a scenario. This enables creating intermediate-year scenarios between two user-defined scenarios.

---

## Current Behavior
  - Interpolates from default start values → original scenario values
  - Uses `input.start_value_for(@scenario)` as the baseline

---

## Proposed Improvement
  - Interpolates from start_scenario values → end_scenario values
  - Validates both scenarios have the same area_code

---

#### 1. Modify Scenario::YearInterpolator Class:
   a. Update the initializer and call method:
      - Add optional start_scenario parameter
      - Store as instance variable
```
  def self.call(scenario, year, current_user = nil, start_scenario = nil)
    new(scenario, year, current_user, start_scenario).run
  end

  def initialize(scenario, year, current_user, start_scenario = nil)
    @scenario = scenario
    @year = year
    @current_user = current_user
    @start_scenario = start_scenario
  end
```   

   b. Update validation to validate start scenario - add to `validate!`, add a `validate_start_scenario!`
   c. Update interpolation logic:
  - Modify interpolate_input_collection to calculate years correctly
  - Modify interpolate_input to use start_scenario values when available ( I think we can fall back to dataset values for now if the start scenario is not given)

#### 2. Modify Api::V3::ScenariosController#interpolate

     a. Accept optional start_scenario_id parameter
     b. Update permitted parameters

#### 3. Tests and documentation

- [ ] Write tests for changes
- [ ] Document changes to the interpolation endpoint

---

## Further enhancements

Outlined above is a MVP for these changes - if we want to go further, we could introduce a batch interpolation endpoint which accepts multiple target years in a single request to make multiple interpolated scenarios at once and enforces a 6 scenario limit.

aaccensi:
## Description

Adds the capability to interpolate between two scenarios with the API endpoint: `POST /api/v3/scenarios/interpolate`
Now if we pass a `start_scenario_id` besides the mandatory `end_year` it will interpolate using the values of the given start scenario instead of the starting values of the scenario targeted by the endpoint.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement
- [ ] Documentation

## Checklist

- [x] I have updated documentation as needed (Work in progress)

## Related Issues

Closes #1684

### DIFF
--- a/app/controllers/api/v3/scenarios_controller.rb
+++ b/app/controllers/api/v3/scenarios_controller.rb
@@ -16,12 +16,21 @@ module Api
         authorize!(:update, Scenario)
       end
 
+      before_action only: %i[interpolate_collection] do
+        # Authorize create here because we load the resources explicity in the action
+        authorize!(:create, Scenario)
+
+        load_batch_scenarios
+      end
+
       before_action only: %i[dashboard merit] do
         authorize!(:read, @scenario)
       end
 
       before_action only: %i[interpolate] do
         authorize!(:clone, @scenario)
+
+        load_start_scenario
       end
 
       before_action only: %i[couple uncouple] do
@@ -207,20 +216,19 @@ module Api
       def interpolate
         result = Scenario::YearInterpolator.call(
           scenario: @scenario,
-          year: params.require(:end_year).to_i,
-          start_scenario_id: params[:start_scenario_id]&.to_i,
-          user: current_user,
-          ability: current_ability
+          year: interpolate_params.require(:end_year).to_i,
+          start_scenario: @start_scenario,
+          user: current_user
         )
 
-        result.either(  
-          lambda { |scenario|  
-            Scenario.transaction { scenario.save! } 
+        result.either(
+          lambda { |scenario|
+            Scenario.transaction { scenario.save! }
             render json: ScenarioSerializer.new(self, scenario)
-          },  
-          lambda { |errors|  
-            render json: { errors: errors.values.flatten }, status: :unprocessable_content  
-          }  
+          },
+          lambda { |errors|
+            render json: { errors: errors.values.flatten }, status: :unprocessable_content
+          }
         )
       rescue ActionController::ParameterMissing
         render json: { errors: ['Interpolated scenario must have an end year'] },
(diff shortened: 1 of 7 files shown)


spec/models/scenario/year_interpolator_spec.rb
```diff
@@ -209,4 +209,38 @@
...
+
+    let(:interpolated) { described_class.call(source, 2040, start_scenario) }
+
+    it 'interpolates based on the start scenario values' do
+      # 50 -> 75 in 20 years
+      # = 62.5 in 10 years
+      expect(interpolated.user_values['grouped_input_one'])
+        .to be_within(1e-2).of(62.5)
+    end
+
```
noracato: Let's add one more test: 
what if an input was set in the scenario, but not in the start_scenario?

I can see from the code that the test would pass, but good to add it for future improvements :)

noracato (commented):
Nice work! Especially on the specs! Thank you for adding the extra spec for the API as well.

It works well as an extension of the existing code. I do think we can win some space by replacing the now many validators by introducing simple `Dry` validation rules. @louispt1 has been working with them a lot lately, he should be able to help you out.

Again well done!

app/controllers/api/v3/scenarios_controller.rb
```diff
@@ -207,22 +203,57 @@ def create
...
+          @scenario,
+          params.require(:end_year).to_i,
+          params[:start_scenario_id]&.to_i,
+          current_user,
+          current_ability
         )
 
-        Scenario.transaction do
-          @interpolated.save!
+        case result
```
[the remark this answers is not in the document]
aaccensi: Oh, I see! This actually fails on the `interpolate_collection` instead of here, but I get your point and I will change it to this style on both endpoints for consistency. Thanks!

app/models/scenario/batch_year_interpolator.rb
```diff
@@ -0,0 +1,129 @@
...
+    result = Contract.new.call(
+      scenario_ids: @scenario_ids,
+      end_years: @end_years
+    )
+
+    result.success? ? Success(nil) : Failure(result.errors.to_h)
+  end
+
+  def fetch_and_validate_scenarios
+    @scenarios = Scenario.where(id: @scenario_ids).to_a
```
noracato: When instantiating a memoized var, it is a good practice to do this in a separate method. Like this, when changing the order of `fetch_and_validate_scenarios` and `validate_target_years` being called will never be a problem.

Also, `.to_a` will force the relation object onto an array. You will loose a lot of optimizations by Rails here. The relation object can do most of what an array can do, and will suffice here.

```ruby
def scenarios
  @scenarios ||= Scenario.accessible_by(@ability).where(id: @scenario_ids).sort_by(&:end_year)
end


def validate_scenarios
  if scenarios.length != @scenario_ids.length
    return Failure(scenario_ids: ["scenarios not found: #{(@scenario_ids - scenarios.map(&:id)).join(', ')}"])
  end
end
```

noracato (changes requested):
The implementation of Dry and the rulesets went well. Thanks for trying it out!

There are a few optimizations to be made regarding Rails. You can check my comments in the code. If you need any more explanations or resources, please ask!

app/models/scenario/batch_year_interpolator.rb
aaccensi: Good one, memoization of scenarios keeps the code more versatile. I'll separate/memoize the scenarios, sort them right away without using to_a. But I think we need to keep the accessibility filtering/validation somehow, either closer as to it is now or a generic fail, or otherwise it could succeed doing something different that what you intended in first place.

app/models/scenario/batch_year_interpolator.rb
```diff
@@ -0,0 +1,129 @@
...
+      if scenario.scaler
+        return Failure(scenario_ids: ["cannot interpolate scaled scenario #{scenario.id}"])
+      end
+      if scenario.start_year != first.start_year
+        return Failure(scenario_ids: ['all scenarios must have the same start year'])
+      end
+      if scenario.area_code != first.area_code
+        return Failure(scenario_ids: ['all scenarios must have the same area code'])
+      end
+    end
```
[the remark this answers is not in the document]
aaccensi: Good to know!

noracato (approved):

louispt1 (approved):

PR #1686, opened 2025-12-29
SavedScenario User Management via Engine

louispt1:
## Description
Implements a pass-through API that proxies SavedScenario user management requests to MyETM.
These changes, along with [this Myetm PR](https://github.com/quintel/my-etm/pull/210) allow users to manage roles on SavedScenarios via the Engine, and solidify the user coupling functionality so accounts are properly coupled based on the callback actions in MyETM.

### Key changes:
  - SavedScenarioUsersController with full CRUD endpoints for bulk user operations (`/api/v3/saved_scenarios/:id/users`)
  - Scenario#copy_roles_from_saved_scenario to replicate user permissions when creating scenarios from presets
  - ETEngine::MyETM module for a pre-configured Faraday client (mimicking setup in MyETM --> We can re-use the `my_etm_client` for other similar Faraday calls)
  - Test coverage for the changes

## Still to do
- Document SavedScenarioUsers CRUD endpoints

Relevant for [this pyetm PR](https://github.com/quintel/pyetm/pull/117)

Closes [#1687](https://github.com/quintel/etengine/issues/1687)

### DIFF
--- a/app/models/scenario.rb
+++ b/app/models/scenario.rb
@@ -363,27 +363,13 @@ class Scenario < ApplicationRecord
     scenario_users.delete_all
   end
 
-  def copy_preset_roles(preset_scenario_users = nil)
+  def copy_preset_roles
     return unless parent
 
-    if preset_scenario_users.present?
-      copy_roles_from_user_data(preset_scenario_users)
-    else
-      copy_roles_from_parent
-    end
+    copy_roles_from_parent
   end
 
   private
-
-  def copy_roles_from_user_data(users_data)
-    users_data.each do |user_data|
-      scenario_users.create(
-        user_id: user_data['user_id'] || user_data[:user_id],
-        user_email: user_data['user_email'] || user_data[:user_email],
-        role_id: user_data['role_id'] || user_data[:role_id]
-      )
-    end
-  end
 
   def copy_roles_from_parent
     parent.scenario_users.each do |preset_user|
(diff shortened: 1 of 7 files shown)


lib/etengine/my_etm.rb
```diff
@@ -0,0 +1,18 @@
...
+    def client
+      @client ||= Faraday.new(url: Settings.identity.issuer) do |conn|
+        conn.request(:json)
+        conn.response(:json)
+        conn.response(:raise_error)
+        conn.options.timeout = 5
+      end
+    end
+  end
+end
```
noracato: Wondering about the use case of interacting with myETM not on behalf of the user.

app/models/scenario.rb
```diff
@@ -369,9 +363,29 @@ def delete_all_users
...
+  end
+
+  private
+
+  def copy_roles_from_user_data(users_data)
+    users_data.each do |user_data|
+      scenario_users.create(
+        user_id: user_data['user_id'] || user_data[:user_id],
+        user_email: user_data['user_email'] || user_data[:user_email],
+        role_id: user_data['role_id'] || user_data[:role_id]
```
noracato: Let's not process params in a model. It would be more elegant to make the `user_data` `with_indifferent_access` before passing them into the `copy_preset_roles` method.

Also wondering if this would not work: 

```ruby
users_data.each { |user_data| scenario_users.create(user_data) }
```

noracato (commented):
Cool stuff! I had just one small doubt about the preset user data. Could you explain your thought process there?

PR #1695, opened 2026-01-15
Remove recursive methods and domestic dead end

### LINKED ISSUE #1617: Retire recursive methods
We identified several unnecessary recursive modules which are no longer required, and some methods which can be refactored, renamed or removed.

Final Demand:
- No longer required, outdated charts and dashboard items should be cleaned up.

DependentSupply:
- Doesn't seem to be used anywhere, remove.

MaxDemand:
- Not used, remove

For the rest there are some question marks still over their usage, but for now no changes are required. The other methods will be re-assessed at a future point The 'abroad' methods should generally be retired, as the model does not use this distinction in the same way anymore.

louispt1:
## Description
Retired FinalDemand, MaxDemand and DependentSupply recursive methods. MaxDemand and DependentSupply could be retired directly, whereas for FinalDemand I had to remove it from a csv output here:
```
app/serializers/application_demands_serializer.rb
```
I also retired the `domestic_dead_end?` method in favour of `right_dead_end?`

Note that the final_demand_group still applies to a lot of nodes in ETSource. This group still seems relevant.

So far everything works well, but our plan to thoroughly test the outcomes by comparing gquery results between local and beta is still in progress.

Closes #1617


aaccensi (approved):
I'd say is good, very hard to tell if anything is left behind but the engine seems to work fine after removing this.

I tested all the dashboard gqueries with py-etm (except dashboard_profitability that did not work for me), before and after the change and they spit the same results!

louispt1:
> The PR looks good, so this is an approval. Just one question before you hit merge.
> 
> Did you also double check the collections? I know they have some final demand charts going on for instance.

The collections just refer to the static final demand group assignments, not the recursive methods so I think it's all good!

PR #1727, opened 2026-03-31
Emissions GQL, load and cache from etsource

noracato:
Introduces a new `Qernel::Emissions` class, containing hashed keys for getting and setting emissions from the new `emissions.csv`, for each dataset in ETSource.

It mimicks the behaviour of `Qernel::Area` in the sense of `area_attributes`. However, emissions can be queried by a sector scope.

Examples

``` ruby
# Get 1990 emissions
EMISSIONS(households, co2, 1990) => 2506777
# Get start year emissions
EMISSIONS('industry.metal', ghg)     => 20778881

# Set value for ghg for a sector (not possible for 1990)
UPDATE(EMISSION(households), ghg, VALUE )
# Set value for co2 for a sector with a subsector (not possible for 1990)
UPDATE(EMISSION('industry.metal'), co2, VALUE )
```

Still to do:
- specs are missing for `Qernel::Emissions`


noracato:
Let's discuss this as well.

PR #1733, opened 2026-04-08
Verify database set up or throw warning

louispt1:
#### Context
Adds a warning when the databases do not set up properly using the databases instructions based on MyETM's readme.

#### Related
Goes with [this myetm PR](https://github.com/quintel/my-etm/pull/238)
Goes with [this etmodel PR](https://github.com/quintel/etmodel/pull/4696)

## Checklist

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -25,6 +25,18 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Preparing database =="
   system! "bin/rails db:prepare"
 
+  puts "\n== Verifying database setup =="
+  verify_cmd = <<~BASH
+    mysql -u root -e "SELECT COUNT(*) FROM scenarios LIMIT 1;" etengine_development 2>&1 > /dev/null
+  BASH
+
+  if system(verify_cmd)
+    puts "✓ Database setup verified successfully"
+  else
+    puts "⚠ Warning: Database may not be fully set up. The scenarios table might be missing."
+    puts "Try running: bin/rails db:drop db:create db:schema:load db:seed"
+  end
+
   puts "\n== Removing old logs and tempfiles =="
   system! "bin/rails log:clear tmp:clear"
 


bin/setup
```diff
@@ -25,6 +25,18 @@ FileUtils.chdir APP_ROOT do
...
+    mysql -u root -e "SELECT COUNT(*) FROM scenarios LIMIT 1;" etengine_development 2>&1 > /dev/null
+  BASH
+
+  if system(verify_cmd)
+    puts "✓ Database setup verified successfully"
+  else
+    puts "⚠ Warning: Database may not be fully set up. The scenarios table might be missing."
+    puts "Try running: bin/rails db:drop db:create db:schema:load db:seed"
+  end
+
```
aaccensi, line 39: If I understand the intent of these changes correctly, I believe this might be not the right check.

Current check only fails on completely missing table, we might be better checking a few more things, like that the migrations have run or that there is the necessary seed data. Also this assumes that mysql is installed even though that is sort of a hard requisite.

The following is not necessary the right solution, but is to illustrate where I'm going with the "right check":
```
puts "\n== Verifying database setup =="
if system("bin/rails runner 'ActiveRecord::Migration.check_pending!' 2>/dev/null")
  puts "✓ Database migrations are up to date"
else
  puts "⚠ Warning: Database may not be fully set up. Migrations may be pending or the schema is missing."
  puts "Try running: bin/rails db:drop db:create db:schema:load db:seed"
end

puts "\n== Verifying seed data =="
result = `bin/rails runner "puts User.where(admin: true).exists?" 2>/dev/null`.strip
if result == "true"
  puts "✓ Seed data verified successfully"
elsif result == "false"
  puts "⚠ Warning: No admin user found. Seeds may not have run."
  puts "Try running: bin/rails db:seed"
else
  puts "⚠ Warning: Could not verify seed data (database may not be reachable)."
end

```

aaccensi (commented):
I think the intention is good but perhaps we need to bring it a step further, let me know what you think?

bin/setup
louispt1: `db:prepare` won't work without mysql, so you should get a more extreme error/warning for that. The seed data is unreliable and often the migrations do not run correctly (for example because there are no scenarios in the case of etengine or no data at all in the case of etmodel). 

I do think we could make this more robust by improving the seed data or being more explicit with the commands (running schema:load for example) but the point of these PRs was more a quick fix for something I saw popping up when new internal users were setting up their DBs, which is that sometimes for etmodel/etengine the databases were not created properly. Sorry that that wasn't clear in the PR text!

louispt1: If you think we should address this stuff more concretely let's put it on the sprint, but in the interest of MVPing it I would propose we just keep it simple. It's also a pain to test these changes as you need to drop your DB and start over 😅

aaccensi (approved):
Alright, as a quick fix sounds good!

PR #1734, opened 2026-04-08
Migration for dispatchable gas plants and update scheme

claudiavalkenier:
#### Context
For a certain project, the goal is to introduce two must-run gas-fired electricity plants, based on existing dispatchable gas-fired electricity plants. These are the CCGT gas plant and the gas turbine plant.

This means that the current CCGT gas plant and gas turbine plants will be renamed to dispatchable, including all related dependencies. Subsequently, duplicates of both plant types will be created and redefined as must-run.

Initially, this will only be implemented for the future year, meaning that no additional data needs to be read from the dataset.

#### Implemented changes
- Migration (for changed input names)

#### Related
- ETModel: https://github.com/quintel/etmodel/pull/4698
- ETSource: https://github.com/quintel/etsource/pull/3433
- ETEngine: https://github.com/quintel/etengine/pull/1734
- Documentation: https://github.com/quintel/documentation/pull/306
- Mechanical Turk: https://github.com/quintel/mechanical_turk/pull/217
- Multi-year-charts: https://github.com/quintel/multi-year-charts/pull/119
- ETDataset: https://github.com/quintel/etdataset/pull/1083

#### Review
Functional reviewer: @louispt1 
- The following (screenshot) came up during the migration. The migration itself was successful, but it may still need to be checked? 
image

Also tagging @kndehaan here!

## Checklist

### DIFF
--- a/db/migrate/20260407120000_rename_gas_plant_inputs.rb
+++ b/db/migrate/20260407120000_rename_gas_plant_inputs.rb
@@ -1,0 +1,21 @@
+require 'etengine/scenario_migration'
+
+class RenameGasPlantInputs < ActiveRecord::Migration[7.0]
+  include ETEngine::ScenarioMigration
+
+  KEYS = {
+    'capacity_of_energy_power_combined_cycle_network_gas' => 'capacity_of_energy_power_combined_cycle_network_gas_dispatchable',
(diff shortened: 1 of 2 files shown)


louispt1 (approved):
Looks great - the error that came up with the logging seems like something to do with the recent ruby 4 / rails 8 upgrade. Probably a few keys are not encoded properly and it wasn't showing up until now.

kndehaan (approved):

PR #1740, opened 2026-05-04
Add Grafana deploy annotation to production pipeline

aaccensi:
#### Context

We recently instrumented our servers with Grafana. It would be helpful to use Grafana annotatations to visually parse what stats period falls within a specific release.

#### Implemented changes

At the end of the Semaphore deploy task for production, add the corresponding Grafana API call to create an annotation to identify the deploy moment. That should appear in the production dashboard graphs similarly to the following test:

image

## Checklist

- [x] I have tested these changes (To some extent, I haven't triggered a deploy though)


noracato (approved):
Cool!

PR #1746, opened 2026-05-11
Fix N+1 cache queries in CommitsController#index

### LINKED ISSUE #1743: N+1 Cache Queries in Etsource::CommitsController#index
#### Context

The commits index page (/etsource/commits) is triggering N+1 Queries since moving to Solid Cache which made apparent that the caching was not optimal. Whether we move back to Memcache or stay in Solid Cache an optimization is warranted so the fix proposed below is cache-backend agnostic.

**Root cause**

The view iterates over all commits and for each one:

Reads a fragment cache key from solid_cache_entries (always a miss on cold cache)
Renders the partial
Writes the fragment back to solid_cache_entries
Reads etsi_semaphore twice via can_import? / import_in_progress? (always a miss on LocalStore, falling through to SolidCache each time)
This produces 4 cache operations per commit (each firing an individual SQL query now). With a full commit list this adds up quickly.

Additionally, the old fragment cache was semantically incorrect: the cached block included the action buttons (Import / Please wait / Loaded), which depend on runtime state. This meant a cached row could show a stale action to the user until the fragment expired.

#### Proposed solution

Two changes required:

Memoize etsi_semaphore at the controller level so it is resolved once per request instead of once per commit. This is a one-liner in import_in_progress? in Etsource::CommitsController.

Cache only the static content by adding a cache_key method to Etsource::Commit (trivial, since sha is already available) and moving the action buttons outside the cache block. Only the _commit partial — which renders static git data (sha, author, message, date) that never changes — is wrapped in cache(commit). The action buttons render directly in the loop, which is now free since import_in_progress? is memoized

aaccensi:
#### Context

The commits index page (/etsource/commits) is triggering N+1 Queries since moving to Solid Cache which made apparent that the caching was not optimal. Whether we move back to Memcache or stay in Solid Cache an optimization is warranted so the fix implemented is cache-backend agnostic.

#### Implemented changes

Two fixes:
- Memoize the etsi_semaphore result in `import_in_progress?` so it is
read from cache once per request instead of once per commit.
- Move action buttons outside the fragment cache block (they depend on
runtime state and should not be cached) and cache only the _commit
partial, which renders static git data that never changes. For this we also added the rails standard method
`cache_key` to `Etsource::Commit` to support `cache(commit)` in the view.

#### Related

Closes #1743

## Checklist

### DIFF
--- a/app/models/etsource/commit.rb
+++ b/app/models/etsource/commit.rb
@@ -102,6 +102,10 @@ module Etsource
       @gcommit.sha
     end
 
+    def cache_key
+      "etsource/commit/#{sha}"
+    end
+
     #######
     private
     #######
--- a/app/controllers/etsource/commits_controller.rb
+++ b/app/controllers/etsource/commits_controller.rb
@@ -160,7 +160,9 @@ class Etsource::CommitsController < ApplicationController
 
   # Returns if an import is currently being performed.
   def import_in_progress?
+    return @import_in_progress unless @import_in_progress.nil?
+
     semaphore = Rails.cache.read(:etsi_semaphore)
-    semaphore && semaphore >= 3.minutes.ago
+    @import_in_progress = semaphore && semaphore >= 3.minutes.ago
   end
 end
--- a/app/views/etsource/commits/index.html.haml
+++ b/app/views/etsource/commits/index.html.haml
@@ -46,15 +46,15 @@
           = date.to_formatted_s(:long)
         - latest_sha = @etsource.get_latest_import_sha
         - commits.each do |commit|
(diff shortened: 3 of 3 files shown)


app/models/etsource/commit.rb
```diff
@@ -102,6 +102,10 @@ def sha
       @gcommit.sha
     end
 
+    def cache_key
```
louispt1, line 105: Had to look into this a bit - nice change!

louispt1 (approved):
LGTM - good stuff

noracato (approved):
Nice changes!

PR #1748, opened 2026-05-11
Use NastyCache for MeritOrder and Molecules to reduce cache round-trips

### LINKED ISSUE #1747: N+1 Cache Queries During GQL Initialization
#### Context

Since adopting Solid Cache, several controller actions that trigger a GQL calculation are flagged by Sentry as N+1 queries. This was not visible before because Memcached round-trips are cheap and don't use SQL; with Solid Cache every cache operation is a SQL query. The fix proposed below is cache-backend agnostic.

During GQL initialization, several etsource modules fetch their static data from `Rails.cache` individually. Each call produces a separate SQL query against `solid_cache_entries`:

- Fever → fever_data
- Reconciliation → reconciliation_hash, reconciliation_carriers
- Molecules → molecules.from_energy_keys, molecules.from_molecules_keys
- MeritOrder → merit_order_hash, hydrogen_hash, heat_network_lt_hash, heat_network_mt_hash, heat_network_ht_hash, agriculture_heat_hash

This data is entirely static between ETSource imports, yet it is fetched with individual Rails.cache.fetch calls on every request, producing ~11 SQL queries per GQL calculation.

#### Proposed solution

`MeritOrder` and `Molecules` are the two classes where batching is practical, they have multiple independent keys with uniform compute logic. `Fever` has a single key so there is nothing to batch. `Reconciliation` has a dependency between its two keys (`reconciliation_hash` depends on `reconciliation_carriers`) so they cannot be fetched in parallel.

For `MeritOrder` and `Molecules`, the individual `Rails.cache.fetch` calls are replaced with a single `Rails.cache.fetch_multi` that reads all keys for that class in one SQL query, computing and writing only missing keys. `Thread.current` is used to ensure `fetch_multi` runs at most once per request, since each `MeritOrder` caller instantiates a fresh object and `Molecules` methods may be called independently multiple times. Subsequent individual `fetch` calls within the same request hit the `LocalStore` (Rails' in-request memory layer populated by `fetch_multi`) at no additional cost.

aaccensi:
#### Context

Since adopting Solid Cache, several controller actions that trigger a GQL calculation are flagged by Sentry as N+1 queries. This was not visible before because Memcached round-trips are cheap; with Solid Cache every cache operation is a SQL query.

#### Implemented changes

`MeritOrder` and `Molecules` previously used `Rails.cache.fetch` on every access. Since this data is purely derived from ETSource and never changes between imports, it is a better fit for `NastyCache`, which stores values in process memory and avoids a cache round-trip on every access.

`NastyCache` already handles invalidation correctly across all Puma workers: on ETSource import, `expire_local!` clears `@cache_store` on the importing process and `mark_expired!` writes a new timestamp to `Rails.cache`. Every other process detects the timestamp change on its next request via `initialize_request` and clears its own store.

#### Related

Closes quintel/etengine#1747

## Checklist

- [X] I have tagged the relevant people for review

### DIFF
--- a/app/models/etsource/merit_order.rb
+++ b/app/models/etsource/merit_order.rb
@@ -1,5 +1,14 @@
 module Etsource
   class MeritOrder
+    CACHE_ATTRIBUTES = %i[
+      merit_order
+      hydrogen
+      heat_network_lt
+      heat_network_mt
+      heat_network_ht
+      agriculture_heat
+    ].freeze
+
     def initialize(etsource = Etsource::Base.instance)
       @etsource = etsource
     end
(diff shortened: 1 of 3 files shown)


louispt1:
I've looked at this a couple of times but I am not confident about it without more investigation.
This cache means that the merit order is preserved for the lifetime of a rails thread, I guess I'm concerned the cache won't be invalidated correctly? But not sure how to test/verify this. I see that [this gem](https://github.com/steveklabnik/request_store) notes the issue and provides a solution, but I am not entirely sure we will even face the issue to begin with. I think we may need to test this carefully

I also see as I was typing this you've updated the cache clearance @aaccensi so maybe all moot now - will take a look later on

aaccensi:
I was looking at the situation with stale data that we should not get after an ETSource import that the solid cache investigation flagged and decided to try to factor also this in here as a test making it more optimal.

louispt1:
> I've looked at this a couple of times but I am not confident about it without more investigation. This cache means that the merit order is preserved for the lifetime of a rails thread, I guess I'm concerned the cache won't be invalidated correctly? But not sure how to test/verify this. I see that [this gem](https://github.com/steveklabnik/request_store) notes the issue and provides a solution, but I am not entirely sure we will even face the issue to begin with. I think we may need to test this carefully
> 
> I also see as I was typing this you've updated the cache clearance @aaccensi so maybe all moot now - will take a look later on

Discussed with @aaccensi and I understand the approach now - I was a little confused before but the approach makes sense to me now and seems practical!

aaccensi:
@noracato , @louispt1 I realized the direction I was taking to improve the caching on these 2 classes was in essence almost using NastyCache, which we already have and it is in fact a minimal change; hence the pivot of this PR.

noracato (approved):

PR #1751, opened 2026-05-12
Block crawlers from the inspect pages and handle missing datasets 

### LINKED ISSUE #1749: Resolve Atlas::DocumentNotFoundError for "de" scenarios on Beta
#### Context
We are receiving a lot of persistent https errors on staging/beta as a result of some "de" scenarios triggering:
`Atlas::DocumentNotFoundError (Could not find a dataset with the key "de")`
On Tuesday (12 May) this error was triggered 98 times in the space of 3 hours (last 24 hours shows 474 errors).

#### Proposed solution

- Removing all these outdated "de" scenarios will resolve the issue at least temporarily 
- The error should be a 404 not a 500 for a missing dataset 
- It seems unlikely that "de" scenarios are being accessed this often. Perhaps adding a robots.txt / some guards against webcrawlers (especially in the engine) could resolve part of the issue 
- In Inspect::LazyGQL catch the error more gracefully and redirect to a working default scenario

louispt1:
#### Context
See [the issue](https://github.com/quintel/etengine/issues/1749) - the high http error rate warning on Grafana is being triggered often, likely due to webcrawlers accessing a scenario with area code "de", triggering the error:
`Atlas::DocumentNotFoundError (Could not find a dataset with the key "de")  `

#### Implemented changes
- Modified lazy_gql and gqueries controller to handle the Atlas::DocumentNotFoundError more gracefully, raise a relevant exception when a dataset is not found and redirect to the home of the inspect controller. The error should not bubble up to the Grafana/Sentry level now, and scenario/user retention policy should handle the outdated scenarios in the future.
- Modified robots.txt and application.html to block webcrawlers from indexing the inspect pages or accessing any inspect paths. There's no benefit I can see from allowing webcrawlers to access these pages.

#### Related
Closes #1749

### DIFF
--- a/app/controllers/inspect/lazy_gql.rb
+++ b/app/controllers/inspect/lazy_gql.rb
@@ -3,18 +3,36 @@ module Inspect
   #
   # Prevents calculating a scenario when it isn't needed.
   class LazyGql
+    class DatasetNotFoundError < StandardError; end
+
     def initialize(scenario)
       @scenario = scenario
       @gql = nil
     end
 
     def method_missing(name, *args, &block)
-      @gql ||= @scenario.gql(prepare: true)
+      @gql ||= initialize_gql
       @gql.public_send(name, *args, &block)
     end
 
-    def respond_to_missing?(method_name)
-      Gql::Gql.public_instance_methods.include?(method_name)
+    def respond_to_missing?(method_name, include_private = false)
+      Gql::Gql.public_instance_methods.include?(method_name) || super
+    end
+
+    private
+
+    def initialize_gql
+      @scenario.gql(prepare: true)
+    rescue Atlas::DocumentNotFoundError, RuntimeError => e
+      raise e unless dataset_not_found_error?(e)
+
+      raise DatasetNotFoundError, "Dataset '#{@scenario.area_code}' not found for scenario #{@scenario.id}"
+    end
+
+    def dataset_not_found_error?(error)
+      error.message.match?(/could not find a dataset with the key/i) ||
+        error.message.match?(/no atlas data for/i) ||
+        (error.message.include?('dataset') && error.message.include?(@scenario.area_code))
     end
   end
 end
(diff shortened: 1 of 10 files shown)


app/controllers/inspect/lazy_gql.rb
```diff
@@ -3,18 +3,36 @@ module Inspect
...
+    rescue Atlas::DocumentNotFoundError, RuntimeError => e
+      raise e unless dataset_not_found_error?(e)
+
+      raise DatasetNotFoundError, "Dataset '#{@scenario.area_code}' not found for scenario #{@scenario.id}"
+    end
+
+    def dataset_not_found_error?(error)
+      error.message.match?(/could not find a dataset with the key/i) ||
+        error.message.match?(/no atlas data for/i) ||
+        (error.message.include?('dataset') && error.message.include?(@scenario.area_code))
```
noracato: We control the errors and error types of the whole chain ourselves. Scanning for messages like this should not be necessary. Try to see if there is a good spot earlier on, where we can introduce a dataset not found error, for instance in `Etsource::Dataset::Import` on line 42.

noracato (commented):
Nice work, thanks for picking this up. I think we can take one more step with the errors.

app/controllers/inspect/lazy_gql.rb
noracato: Or maybe even way more simple in this case is to add a check instead of letting the error bubble up:

`raise DatasetNotFoundError unless Atlas::Dataset.find(@scenario.area_code)` in the initialiser

louispt1: I took a slightly different approach with: https://github.com/quintel/etengine/pull/1751/changes#diff-fb609d8d150083bdc86d217f3a6cf354d2150fe891225c8acbe2ac52daaf551d

How do you feel about this? I agree matching the error messages doesn't feel great

noracato: This feels a lot cleaner to me! I like how all different controllers can now handle the dataset not found in the same way.

noracato (approved):

PR #1753, opened 2026-05-20
Emissions CSV export

aaccensi:
#### Context

As part of the emissions project we want to be able to export the emissions data in a CSV. The existing ConfiguredCSVSerializer serializes graph data to CSV based on a YAML config in ETSource, but it only supported simple query-based columns. The emissions export requires iterating over nodes in a group and reading attributes directly from each node, including support for both the energy and molecules graphs.

#### Implemented changes

Expanded ConfiguredCSVSerializer to support two new column types:
- node_group: A hidden column whose value names a node group. The config row is expanded into one CSV row per node in that group, combining nodes from both the energy and molecules graphs.
- node_attribute: Evaluates a Ruby expression against node_api for each expanded node. Supports an optional transform: expression (with value bound to the result) for post-processing, e.g. unit conversion or conditional mapping.

Other changes:
- Added an optional `period:` parameter (`:present` or `:future`) to the serializer to select which graph to use for node expansion. Existing column types (present, future, unit, query) are unaffected.
- Fixed label: being silently ignored for non-query column types.
- Added `emissions_present` and `emissions_future` CSV endpoints backed by a new `emissions_csv.yml `config in ETSource.

#### Related

This effort is in 3 places but only 2 of them have PRs (to merge back into the general emissions branch)
- etengine: https://github.com/quintel/etengine/pull/1753
- etsource: https://github.com/quintel/etsource/pull/3443
- etmodel: No PR, created emissions branch that didn't exist yet

## Checklist


noracato (approved):
Very elegant solution with the `transform` attribute. Nice find!

Specs look good as well! Great work

kndehaan:
@noracato @aaccensi we updated the `direct_emissions_csv.yml` on ETSource to contain all the relevant groups. Robin also updated the spec fixture file on ETEngine. Does this spec file indeed need to be updated as well or should we undo this commit?

noracato:
ETEngine should generally only be touched by devs. The specs are fast internal tests to see if the code is behaving correctly under different cases and circumstances. I will revert the commit for you.

kndehaan:
Seems to look good! @noracato could you do a final review of Andreu's changes? If approved, we can merge it.

app/models/qernel/node_api/molecule_emissions.rb
```diff
@@ -12,6 +12,14 @@ module NodeApi
     # - CO2 capture = placeholder (returns 0.0 for now)
     # - Total GHG = CO2 production - capture + other GHG
     module MoleculeEmissions
+      # The GHG carrier type for this node: 'Other GHG' for molecule nodes whose input slot
+      # carries other_ghg, 'CO2' for all others.
+      #
+      # @return [String] 'Other GHG' or 'CO2'
+      def ghg_carrier
+        inputs.map { |s| s.carrier&.key }.include?(:other_ghg) ? 'Other GHG' : 'CO2'
```
noracato:
```suggestion
inputs.map { |s| s.carrier&.key }.include?(:other_ghg) ? :other_ghg : :co2
```

It would be good to keep the symbol here instead of the string, so we can reuse the method later in the Qernel.

You are converting all node attributes to strings anyways in the serializer, so they will land well in the CSV. We will lose the capitalisation, but I hope everybody can live with that.

aaccensi: Alright, this was discussed with Kyra and done in the spirit of making the files closer to the user story example and because I imagined the method to be very niche to be reused anywhere. Nonetheless, looking for reusability as unlikely as it is, is indeed the right approach. If there is still a desire to have prettier text in that column we could solve that with the new "transform" directly on the yml file.

PR #1754, opened 2026-05-21
Emissions

louispt1:
Summary of changes

##### Emissions Calculation Framework
- New DirectEmissions module implementing mass-balance equations for fossil and biogenic CO2
- Tracks carbon flows: input content (A) + utilization (B) - output content (C) - capture (D) = emissions (E)
- New MoleculeEmissions module for emissions reporting on molecule nodes
- Support for CO2 capture (via `ccs_capture_rate`) and utilization (via `co2_utilisation_per_mj`)

##### Carbon Content Tracing
- RecursiveFactor::DirectEmissions module for tracing CO2 content through supply chains
- Handles mixed carriers (network gas, crude oil) by recursively calculating weighted composition
- New `emissions_skip_crude_oil_mix` edge group for forcing weighted mix calculations

##### GQL & Dataset Integration
- New EMISSIONS() GQL function for accessing emissions data from datasets
- Emissions data loading infrastructure in `Dataset::Import` and `Etsource::Loader`
- Graph-level emissions hash storage

##### Supporting Infrastructure
- emissions node group for nodes participating in emissions tracking
- ccus_captured node group for CO2 removal/capture technologies
  - Uses lazy memoization for ccus_captured? check to avoid Marshal serialization issues
- Updated biogenic emissions (primary) to use `free_co2_factor` for capture calculations with consistent results

##### Other
- Reporting methods return nil for non-emissions nodes (checked via with_emissions_node)
- Atlas gem updated to support emissions data structures

Note: Atlas reference should be updated once the Atlas `emissions` branch has been merged.
Goes with:
- https://github.com/quintel/etengine/pull/1754
- https://github.com/quintel/atlas/pull/189
- https://github.com/quintel/etsource/pull/3444
- https://github.com/quintel/etmodel/pull/4719
- https://github.com/quintel/etdataset/pull/1087
- https://github.com/quintel/etlocal/pull/695
- https://github.com/quintel/documentation/pull/321


kndehaan:
Note that `direct_co2_output_content_carriers_biogenic` doesn't give the correct results yet, also leading to incorrect results for methods that depend on this one. 

This has still to with the `potential_co2_conversion_per_mj` attribute missing on carriers, as a consequence going to recursion where it should not always be so. This is something that we still need to look into.

louispt1:
@kndehaan I have pushed a fix for the biogenic output content carriers now

app/models/qernel/graph.rb
```diff
@@ -189,6 +191,8 @@ def remove_dataset_attributes
   def refresh_dataset_attributes
     # See Qernel::Dataset#assign_dataset_attributes to understand what's going on:
     call_on_each_qernel_object(:assign_dataset_attributes)
+    # Manually assign emissions hash (not a DatasetAttributes object)
+    @emissions = dataset&.data&.[](:emissions) || {}
```
noracato: Can you explain once more why not to use `DatasetAttributes`? Like this there is no validation on what users can set as an emissions attribute.

spec/models/gql/runtime/functions/update_spec.rb
```diff
@@ -49,4 +49,16 @@
...
+
+  describe 'UPDATE(EMISSIONS(), buildings_non_specified_energetic_other_ghg, 100.0)' do
+    before { result }
+
+    it 'sets the emissions to 100.0' do
+      expect(gql.query_future('EMISSIONS(buildings_non_specified, energetic, other_ghg)')).to eq(100.0)
+    end
+  end
+
 end
```
noracato, line 145: Missing tests for nonexistent keys

noracato (changes requested):
I'm worried that the UPDATE part does not have enough validation.

Can you please add tests that show updating nonexistent keys will not work? I have a feeling we will need the `DatasetAttributes` module after all (it is still on the `emissions-gql` branch)

app/models/qernel/graph.rb
louispt1: Good point, I was not really looking at UPDATE yet, so I thought to keep it as light as possible for read. 
However, you are right that it's cleaner to just use the DatasetAttributes approach for consistency and validation.

I re-instated the DatasetAttributes approach and added more tests to cover the case of nonexistent keys etc. I expect more spec will be needed based on the changes for 1990 but I think it's best to process that separately.

kndehaan:
Adding myself as reviewer to check the descriptions for the methods (it should be understandable and correct from a modeller's perspective as well).

app/models/qernel/node_api/direct_emissions.rb
```diff
@@ -0,0 +1,274 @@
+# frozen_string_literal: true
+
+module Qernel
+  module NodeApi
+    # Calculates direct CO2 emissions from fossil fuels using a mass balance approach.
```
kndehaan:
```suggestion
    # Calculates direct CO2 emissions from fossil and biogenic carriers using a mass balance approach.
```

app/models/qernel/node_api/direct_emissions.rb
```diff
@@ -0,0 +1,274 @@
...
+    #
+    # Direct emissions measure the change in carbon stock at a specific node:
+    # - Carbon enters via input carriers
+    # - Carbon leaves via output carriers
+    # - Direct emissions = CO2 released to atmosphere at this location (input - output)
```
kndehaan, line 12:
```suggestion
    # Direct emissions measure the change in carbon stock at a specific node:
    # - Carbon enters via input carriers
    # - Carbon enters via CO2 utilisation
    # - Carbon leaves via output carriers
    # - Carbon leaves via CO2 capture
    # - Direct emissions = CO2 released to atmosphere at this location (input - output)
```

app/models/qernel/node_api/direct_emissions.rb
```diff
@@ -0,0 +1,274 @@
...
+    #
+    # For mixed carriers (network gas, crude oil, carrier_mix), composition tracking
+    # uses RecursiveFactor::WeightedCarrier to recursively trace CO2 content through the supply chain.
+    module DirectEmissions
+      # CO2 content of all carriers entering the node.
+      #
+      # @return [Float, nil] Total CO2 content from input carriers in kg, or nil if node is not in emissions group
```
kndehaan:
```suggestion
      # Fossil CO2 content of all carriers entering the node.
      #
      # @return [Float, nil] Total fossil CO2 content from input carriers in kg, or nil if node is not in emissions group
```

app/models/qernel/node_api/direct_emissions.rb
```diff
@@ -0,0 +1,274 @@
...
+      def direct_co2_input_content_carriers_fossil
+        with_emissions_node do
+          sum_carbon_content(inputs.flat_map(&:edges), :fossil)
+        end
+      end
+
+      # CO2 content of all carriers leaving the node.
+      #
+      # @return [Float, nil] Total CO2 content from output carriers in kg, or nil if node is not in emissions group
```
kndehaan:
```suggestion
      # Fossil CO2 content of all carriers leaving the node.
      #
      # @return [Float, nil] Total fossil CO2 content from output carriers in kg, or nil if node is not in emissions group
```

app/models/qernel/node_api/direct_emissions.rb
```diff
@@ -0,0 +1,274 @@
...
+          sum_carbon_content(output_edges, :fossil)
+        end
+      end
+
+      # CO2 emissions produced at this node.
+      #
+      # Calculates net emissions after accounting for CO2 utilisation.
+      # Emissions = input co2 + utilised co2 - captured co2 - output co2
+      #
+      # @return [Float, nil] Direct fossil CO2 emissions in kg, or nil if node is not in emissions group
```
kndehaan:
```suggestion
      # Fossil CO2 emissions produced at this node.
      #
      # Calculates net emissions after accounting for CO2 utilisation and CO2 capture.
      # Emissions = input carriers co2 + utilised co2 - captured co2 - output carriers co2
      #
      # @return [Float, nil] Direct fossil CO2 emissions in kg, or nil if node is not in emissions group
```

app/models/qernel/node_api/direct_emissions.rb
```diff
@@ -0,0 +1,274 @@
...
+      end
+
+      # Biogenic CO2 emissions produced at this node.
+      #
+      # Emissions = input biogenic co2 - captured biogenic co2 - output biogenic co2
+      #
+      # Note: Unlike fossil emissions, biogenic CO2 utilisation is not included as
+      # feedstock utilisation typically involves fossil CO2.
+      #
+      # @return [Float, nil] Direct biogenic CO2 emissions in kg, or nil if node is not in emissions group
```
kndehaan:
```suggestion
      # Emissions = input carriers biogenic co2 - captured biogenic co2 - output carriers biogenic co2
      #
      # Note: Unlike fossil emissions, biogenic CO2 utilisation is not included as 
      # the model does not distinguish biogenic CO2 utilisation. 
      #
      # @return [Float, nil] Direct biogenic CO2 emissions in kg, or nil if node is not in emissions group
```

app/models/qernel/node_api/direct_emissions.rb
```diff
@@ -0,0 +1,274 @@
...
+            direct_co2_input_content_carriers_biogenic -
+              direct_co2_output_content_carriers_biogenic
+          )
+        end
+      end
+
+      # Fossil CO2 utilised (consumed as feedstock) at this node.
+      #
+      # Calculates CO2 that is consumed as feedstock rather than emitted, based on the total
+      # output energy and the node's co2_utilisation_per_mj attribute.
```
kndehaan:
```suggestion
      # Calculates CO2 that is consumed as feedstock rather than emitted, based on the total
      # useful carrier output in MJ and the node's co2_utilisation_per_mj attribute.
```

app/models/qernel/node_api/direct_emissions.rb
```diff
@@ -0,0 +1,274 @@
...
+      #
+      # Currently returns only fossil utilisation, as biogenic utilisation is always 0.
+      #
+      # @return [Float, nil] Total CO2 utilised in kg, or nil if node is not in emissions group
+      def direct_co2_input_utilisation
+        with_emissions_node do
+          direct_co2_input_utilisation_fossil
+          # Potentially in the future: + direct_co2_input_utilisation_biogenic (currently 0)
+        end
+      end
```
kndehaan: I think we decided to remove this method for now, as we currently don't use it right now, right? @louispt1

kndehaan (commented):
I put up some text suggestions and placed one comment about removing a method. Let me know if you have questions @louispt1

app/models/qernel/node_api/direct_emissions.rb
louispt1: Yes this method snuck back in with the csv serialiser merge - I will remove it

app/models/qernel/emissions.rb
```diff
@@ -0,0 +1,93 @@
...
+    # GQL uses this to scope the sector for easier queries and input/update
+    # statements in etsource
+    #
+    # Example:
+    #   EMISSIONS(households, energetic) returns a ScopedSector
+    #   Then UPDATE can call: scoped.co2 = 100.0
+    class ScopedSector
+      # Valid GHG types that can be used in emission keys
+      VALID_GHG_TYPES = %w[co2 other_ghg n2o ch4 hfc pfc sf6 nf3].freeze
+      VALID_GHG_PATTERN = /^(#{VALID_GHG_TYPES.join('|')})(_\d{4})?$/.freeze
```
noracato: Sorry to comment again, but let's not use constants with carriers and patterns. These keys have nothing to do with engine validations, they are part of the dataset pipeline IMO - or at least ETSource. The engine should not carry this knowledge.

noracato (changes requested):

app/models/qernel/emissions.rb
louispt1: No need to apologise, you're spot on. I pushed a change removing the validation.
Edit:
However now I need to figure out how to handle validation a little smarter - still working on it :)

louispt1: @noracato what do you think about [this approach?](https://github.com/quintel/etengine/pull/1754/commits/b7bce7cef773784126542b06255476d26c71d776)

app/models/qernel/emissions.rb
```diff
@@ -57,7 +64,17 @@ def method_missing(method_name, *args)
...
+        return false if method_name.to_sym == :query
+
+        # For setters, check if the emission key exists in the dataset
+        if method_name.to_s.end_with?('=')
+          data_key = scoped_method(method_name.to_s.sub(/=$/, ''))
+          @emissions.dataset_has_key?(data_key)
+        else
+          # Getters always respond (may return nil if key doesn't exist)
+          true
+        end
```
noracato:
```suggestion
# Getters always respond (may return nil if key doesn't exist)
return true unless method_name.to_s.end_with?('=')

@emissions.dataset_has_key?(
  scoped_method(method_name.to_s.sub(/=$/, ''))
)
```

General approach is good. This is a bit pythonesque ;)

app/models/qernel/emissions.rb
```diff
@@ -73,5 +90,17 @@ def initialize(graph = nil)
...
+    # Public: Check if a key exists in the loaded emissions dataset
+    #
+    # key - String or Symbol representing the emission key
+    #
+    # Returns Boolean
+    def dataset_has_key?(key)
+      return false unless dataset_attributes
+
+      # Check both string and symbol keys since datasets may use either
+      dataset_attributes.key?(key.to_s) || dataset_attributes.key?(key.to_sym)
```
noracato: I don't feel this should be handled here?

app/models/qernel/emissions.rb
noracato: Actually, what was wrong with:

```ruby
def respond_to_missing?(method_name, include_private = false)
  data_key = scoped_method(method_name).split('=').first

  @emissions.respond_to?(data_key) || super
end
```

noracato (changes requested):
Sorry to keep nitpicking. There is still logic present in the module that I feel over complicates it.

noracato (approved):

kndehaan (dismissed):
I found a critical bug that needs to be fixed. It concerns using the EMISSIONS() function for setting `demand` on dataset value molecule nodes. It works for the nl datasets (ETDataset datasets), but I think it does not work as it should for ETLocal datasets. 

Example blank nl2023 scenario, querying the three things below gives the expected, desired other ghg emissions (don't mind the differences in decimals):

```
EACH(
  EMISSIONS(agriculture_non_specified, non_energetic, other_ghg),
  MV(agriculture_non_specified_non_energetic_other_ghg, demand),
  MV(agriculture_non_specified_non_energetic_other_ghg, direct_reporting_emissions_other_ghg_emissions)
)

[
  17,888.11512,
  17,888,115,120.0,
  17,888,115,120.0,
]
```

This is however not the case for for example blank NO_norway scenario (and other countries as well):

```
EACH(
  EMISSIONS(agriculture_non_specified, non_energetic, other_ghg),
  MV(agriculture_non_specified_non_energetic_other_ghg, demand),
  MV(agriculture_non_specified_non_energetic_other_ghg, direct_reporting_emissions_other_ghg_emissions)
)



[
  4,516.46047,
  0.0,
  0.0,
]
```

So I think something in reading with `~ demand` on the molecule nodes goes wrong when it's a ETLocal dataset (derived dataset). 

I know that for energy nodes, the derived dataset first looks if `graph_methods` is specified. If not, it falls back to what is defined with `~`, whereas full datasets directly look at `~` and don't look at `graph_methods`. 

Might it be that for the molecule nodes, something has to be configured for derived datasets to also look at values specified with`~`?

noracato:
I'll have a look. Gut feeling is that you are right. There might also be a hidden scaling factor involved.

kndehaan:
Additionally, if I set a hard-coded value on the molecule node like `~ demand = 1000.0`, querying the demand for this node for Norway returns zero, whereas it does return the hard-coded value for nl2023. This confirms that something doesn't go right for setting demand with `~` method for derived datasets.

noracato:
I found the culprit.

When calculating the initial graph, there is an extra thing applied for `Derived` datasets (ETLocal datasets). It is called `ZeroMoleculeNodes`. And it just sets the demand for all molecule nodes to zero.

[Here](https://github.com/quintel/atlas/blob/f2e3f0de7a7fc9cd864fcf47ed3624c812803e83/lib/atlas/runner/zero_molecule_nodes.rb#L5) you can see a comment saying "Derived datasets have no molecule flows."

I will try to find out why this was setup like this, and if we can lift it, or if we can create a special case for "floating" nodes.

mabijkerk:
In the end, working with the emission methods as they are defined now is not the most intuitive: it is a bit too elaborate in practice. How did you experience this @kndehaan? Not saying this is a must have, but still good to flag.

More straightforward might be:
| current method label | proposed method label |
| --- | --- |
| direct_co2_input_content_carriers_fossil | direct_co2_content_input_fossil |
| direct_co2_input_utilisation_fossil | direct_co2_utilisation_fossil |
| direct_co2_output_content_carriers_fossil | direct_co2_content_output_fossil |
| n.a. | direct_co2_production_fossil |
| direct_co2_output_production_capture_fossil | direct_co2_capture_fossil |
| direct_co2_output_production_emissions_fossil | direct_co2_emissions_fossil |

Also, a `direct_reporting_emissions_co2_emissions` method would be useful. It is not the most intuitive that I have to use the `direct_reporting_emissions_total_ghg_emissions` method to get the complete picture of a node.

kndehaan:
@mabijkerk let's review your comment in the next increment. I agree that the naming of the methods has not been most intuitive. Also, being able to query the direct reporting CO2 emissions of a node would be useful.

PR #1757, opened 2026-06-04
Discard passthrough

louispt1:
#### Context
- Added a discard endpoint on the saved scenarios controller which passes through to the myetm discard action.
- Updated the upsert transition path to handle nested params - we expect nested params everywhere else, and pyetm makes calls with nested params. This unifies the approach.
- Updated the my_etm_client to raise errors so error messages on saved scenario and collection endpoints passthrough from myetm to the API caller of etengine (in this case pyetm)
- Lots of spec

Note: Also updated the rubocop config so I could check my changes for issues

#### Related
Goes with [this myetm PR](https://github.com/quintel/my-etm/pull/269)

## Checklist

- [ ] I have updated documentation as needed --> Docs still need to be updated

### DIFF
--- a/app/controllers/api/v3/base_controller.rb
+++ b/app/controllers/api/v3/base_controller.rb
@@ -88,6 +88,7 @@ module Api
       def my_etm_client
         Faraday.new(url: Settings.identity.issuer) do |conn|
           conn.response(:json)
+          conn.response(:raise_error)
           conn.request(:json)
 
           if (auth_header = request.authorization.to_s[/\ABearer (.+)\z/, 1])
--- a/app/controllers/api/v3/base_controller.rb
+++ b/app/controllers/api/v3/base_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Api
   module V3
     class BaseController < ActionController::API
--- a/app/controllers/api/v3/saved_scenarios_controller.rb
+++ b/app/controllers/api/v3/saved_scenarios_controller.rb
@@ -73,6 +73,18 @@ module Api
         render_not_found
       end
 
+      def discard
+        DiscardSavedScenario.new.call(
+          id: params.require(:id),
+          client: my_etm_client
+        ).either(
+          ->((data, *)) { render json: data },
+          ->(error)     { service_error_response(error) }
+        )
+      rescue Faraday::ResourceNotFound
+        render_not_found
+      end
+
       private
 
       def hydrate_scenarios(saved_scenarios)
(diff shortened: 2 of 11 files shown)


app/controllers/api/v3/base_controller.rb
```diff
@@ -88,6 +90,7 @@ def render_not_found(body = { errors: ['Not found'] })
       def my_etm_client
         Faraday.new(url: Settings.identity.issuer) do |conn|
           conn.response(:json)
+          conn.response(:raise_error)
```
noracato, line 93: Double check: will this break other workflows?

app/controllers/api/v3/saved_scenarios_controller.rb
```diff
@@ -73,14 +77,26 @@ def destroy
...
+
       private
 
       def hydrate_scenarios(saved_scenarios)
         saved_scenarios = [saved_scenarios] unless saved_scenarios.is_a?(Array)
 
         scenarios = Scenario
           .accessible_by(current_ability)
-          .where(id: saved_scenarios.map { |s| s['scenario_id'] })
+          .where(id: saved_scenarios.pluck('scenario_id'))
```
noracato, line 103: Nice catch

app/controllers/api/v3/saved_scenarios_controller.rb
```diff
@@ -114,7 +130,7 @@ def update_pagination_links(links)
       end
 
       def saved_scenario_params
-        params.require(:saved_scenario).permit(:scenario_id, :title, :description, :private)
+        params.expect(saved_scenario: %i[scenario_id title description private])
```
noracato, line 137: Why did you move from require to expect?

noracato (approved):

app/controllers/api/v3/saved_scenarios_controller.rb
louispt1: Rubocop did this via autocorrect - it is a Rails 8 change: https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect

I think it's quite neat! But can also revert to require if we only want to update this kind of thing 'all at once' rather than incrementally like this

noracato: No super cool! Nice to know!

app/controllers/api/v3/base_controller.rb
louispt1: Good one - I had a look around and added [this commit](https://github.com/quintel/etengine/pull/1757/commits/303d2c9e163f3d7114d827cef01d880f5ad57435) to protect from issues. I think now this is ok?

noracato (approved):

PR #1759, opened 2026-06-04
Emission validation

noracato:
### Context

For the emissions project we want to validate node methods against real dataset data.
I created a small lib module a la mechanical turk, which can run against your local ETSource. The idea is that this module can once be extracted towards a validation gem.

Effort spent is minimal. The config for instance was the easiest way now, but in the future we might want to add filters instead (eg `group: country`). As such there are many improvements. These are just quick bare bones.

### Implemented changes

In the file `lib/graph_data_validation` I created a new rspec module able to load graphs based on datasets and test node methods.

In the `config.yml` you can specify different groups. The group you suggested now is called countries.

You can see in the `spec/datasets_spec.rb` that for each country, the emission validation will be run (`include_examples 'emissions', country`).

The emission validation can be found in `spec/validations/emissions.rb`. 

You can run the specs by being in etengine in the terminal and running:

`bundle exec rspec lib/graph_data_validation/spec`

It will run against your local ETSource, and will print the dataset it is working on (as it takes a long time). But will only print all validation mistakes when it is done with all of them.

Screenshot 2026-06-04 at 16 03 20

## Checklist


kndehaan (approved):

lib/graph_data_validation/spec/spec_helper.rb
```diff
@@ -0,0 +1,23 @@
+# Load module code
+puts "HELLO"
```
louispt1:
```suggestion

```

lib/graph_data_validation/spec/spec_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+# Load examples
+Dir["./spec/validation/**/*.rb"].sort.each { |f| require f }
+
+RSpec::Matchers.define :match_output do |output|
+  match do |input|
+    margin = 1.0E-12 * input
+    input - output >= -margin
+  end
+
+  failure_message_for_should do |input|
```
louispt1:
```suggestion
  failure_message do |input|
```
[this one is deprecated](https://www.rubydoc.info/github/rspec/rspec-expectations/RSpec/Matchers/DSL/Macros/Deprecated:failure_message_for_should)

louispt1 (approved):
Looks good, works well 👍 

Excited to gemmify it one day

lib/graph_data_validation/spec/spec_helper.rb
noracato: Thanks, I copied it straight from MT. I'll make an issue there as well that we should update it!

PR #1762, opened 2026-06-09
Add node copy key without styling

mabijkerk:
#### Context
I often copy nodes from the node view to Excel, but then it takes the formatting along which makes the sizing very big.

image

#### Implemented changes
I had AI add a button next to the buttons that are already there that allows you to quickly copy the node name without styling.

image

image

Note: if do not know if this matches with the preferred implemtentation in Ruby for the devs.

## Checklist


louispt1:
I made some edits, changing the button to sit just next to the node name:
image

When you copy there is a visual indicator to let you know the node name is added to your clipboard:
image

I re-used some css from the copy function that had been implemented for the staff applications, which have since moved to MyETM. As a result I also noticed that there was some dead staff applications code lying around that needed to be cleaned up. I've added that in [a separate PR](https://github.com/quintel/etengine/pull/1763).

In the end I tried to process this as quickly as possible, but I still ended up effectively 'doing' the task.

I have tagged @noracato for review

noracato (approved):

mabijkerk:
Very cool, exactly what I had in mind!

Only minor detail is that the icon is not neatly aligned in terms of size with the node name. Finally, what if the node name is very long, does the copy icon get pushed out? Should the copy icon be at the 'beginning'?

louispt1:
This is a bit more the same size:
image

and now node names (and the copy icon) wrap when the window is smaller:
image

mabijkerk:
Dope! Let's merge?

PR #1768, opened 2026-06-17
Capacity data exports

noracato:
See #1767


mabijkerk:
@aaccensi it seems like the data is rounded at 4 decimals, is that right? Can we set it to 6?

aaccensi:
Sure, just so you know I used 4 because the curves were using 4 so it seemed a sensible parallel.

mabijkerk:
@aaccensi your earlier explanation about storage does not make sense to me entirely. Yes, the input and output capacities are different, but the installed **input** capacity should not exceed the peak **input** capacity. Nor should the installed **output** capacity exceed the peak **output** capacity.

For example for energy_hydrogen_storage_salt_cavern in II3050v2 DEC I'm seeing something strange:
key | installed_capacity | peak_capacity
-- | -- | --
energy_hydrogen_storage_salt_cavern.output   (MW) | 45031.162 | 51286.8249

I cannot reproduce how this peak output capacity is caculated. Can you?

config/routes.rb
```diff
@@ -102,16 +101,24 @@
...
-            get :hydrogen,               to: 'curves#hydrogen',               as: :hydrogen_download
-            get :network_gas,            to: 'curves#network_gas',            as: :network_gas_download
-            get :residual_load,          to: 'curves#residual_load',          as: :residual_load_download
-            get :hydrogen_integral_cost, to: 'curves#hydrogen_integral_cost', as: :hydrogen_integral_cost_download
+            get :buildings_heat,          to: 'curves#buildings_heat_curves',   as: :buildings_heat_download
+            get :merit_order,             to: 'curves#electricity_profiles'     # backwards compatible
+            get :electricity_profiles,    to: 'curves#electricity_profiles',    as: :electricity_profiles_download
+            get :electricity_price,       to: 'curves#electricity_price',       as: :electricity_price_download
+            get :heat_network,            to: 'curves#heat_network_profiles'    # backwards compatible
+            get :heat_network_profiles,   to: 'curves#heat_network_profiles',   as: :heat_network_profiles_download
```
mabijkerk: All the data exports are named after their carriers, should we name this `steam_hot_water` instead of `heat_network`?

mabijkerk:
@aaccensi I cannot reproduce how the cooling technologies take their installed capacity, for example in II3050v2 DEC for airconditioning I don't know how the installed capacity is calculated:

key | installed_capacity | peak_capacity
-- | -- | --
households_cooling_airconditioning_electricity.input (MW) | 1207.5313 | 157.1077

config/routes.rb
noracato: I believe agriculture heat is not included in this download, but does contain that carrier. So the name would not really cover the download

aaccensi:
> For example for energy_hydrogen_storage_salt_cavern in II3050v2 DEC I'm seeing something strange:
> 
> key	installed_capacity	peak_capacity
> energy_hydrogen_storage_salt_cavern.output (MW)	45031.162	51286.8249
> I cannot reproduce how this peak output capacity is caculated. Can you?

Relevant queries:
```
																	2019		2050
																	--------	--------
V(energy_hydrogen_storage_salt_cavern, number_of_units)				0.0			85.67998137463711
V(energy_hydrogen_storage_salt_cavern, input_capacity)				1,000.0 	598.5858553339008
V(energy_hydrogen_storage_salt_cavern, typical_input_capacity)		1,000.0		598.5858553339008
V(energy_hydrogen_storage_salt_cavern, hydrogen_output_capacity)	1,000.0		598.5858553339008
V(energy_hydrogen_storage_salt_cavern, hydrogen_output_conversion)	1.0			0.8780259267083321
```
So according to the installed_capacity formula used on the export I believe we can see what is going on in this case:
`installed_capacity = hydrogen_output_conversion × input_capacity × number_of_units`:
- 1.0 × 598.5858553339008 × 85.67998137463711 = 51286.8249361
- 0.8780259267083321 × 598.5858553339008 × 85.67998137463711 = 45031.1619925

Here we can see that `hydrogen_output_conversion` is 0.878 at runtime (storage loss?), not the 1.0 from etsource. `peak_capacity` is just `output_curve.max`, which is unbounded by the rating.

aaccensi:
> @aaccensi I cannot reproduce how the cooling technologies take their installed capacity, for example in II3050v2 DEC for airconditioning I don't know how the installed capacity is calculated:
> 
> key	installed_capacity	peak_capacity
> households_cooling_airconditioning_electricity.input (MW)	1207.5313	157.1077

Relevant queries:
```
																					2019				2050
																					--------			--------
V(households_cooling_airconditioning_electricity, number_of_units)					554,360.4179410399	1,932,050.068014174
V(households_cooling_airconditioning_electricity, input_capacity)					0.0025				0.0025
V(households_cooling_airconditioning_electricity, typical_input_capacity)			0.0					0.0
V(households_cooling_airconditioning_electricity, heat_output_capacity)				0.0025				0.0025
V(households_cooling_airconditioning_electricity, cooling_based_input_capacity)		0.0025				0.0025
V(households_cooling_airconditioning_electricity, electricity_input_conversion)		0.25				0.25
```

So according to the installed_capacity formula used on the export I believe we can see what is going on in this case:
`installed_capacity = electricity_input_conversion × input_capacity × number_of_units`
- 0.25 × 0.0025 × 1,932,050 = 1207.53


In here our `input_capacity` method fallback strategy reaches `cooling_based_input_capacity` because there is no `typical_input_capacity`. In the code (cost.rb:697-703) we can see this value comes from the the following calculation:
`cooling_based_input_capacity = heat_output_capacity / cooling_output_conversion `:
- 0.0025 / 1.0 = 0.0025

mabijkerk:
> Here we can see that hydrogen_output_conversion is 0.878 at runtime (storage loss?), not the 1.0 from etsource. peak_capacity is just output_curve.max, which is unbounded by the rating.

Right, this is the unused hydrogen in storage at the end of the calculation year (31st of March). This gets put on the loss edge to the `energy_hydrogen_storage_unused` node. In this case the conversion efficiency is incorreclty applied to correct capacity. Thanks for the investigation, I'll add it to the list of debts.

> In here our input_capacity method fallback strategy reaches cooling_based_input_capacity because there is no typical_input_capacity.

Clear, I'll add it to the list of debts.

config/routes.rb
mabijkerk: Then shouldn't we call it `district_heating`? We have the industrial steam network, the local agriculture CHP network and the district heating network (LT/MT/HT). They are all heat networks.

noracato: Makes sense to me! 👍

mabijkerk (approved):

config/routes.rb
```diff
@@ -102,16 +105,20 @@
...
-            get :household_heat,         to: 'curves#household_heat_curves',  as: :household_heat_download
-            get :hydrogen,               to: 'curves#hydrogen',               as: :hydrogen_download
-            get :network_gas,            to: 'curves#network_gas',            as: :network_gas_download
-            get :residual_load,          to: 'curves#residual_load',          as: :residual_load_download
-            get :hydrogen_integral_cost, to: 'curves#hydrogen_integral_cost', as: :hydrogen_integral_cost_download
+            get :buildings_heat,            to: 'curves#buildings_heat_curves',     as: :buildings_heat_download
+            get :merit_order,               to: 'curves#electricity_profiles'       # backwards compatible
+            get :electricity_profiles,      to: 'curves#electricity_profiles',      as: :electricity_profiles_download
+            get :electricity_price,         to: 'curves#electricity_price',         as: :electricity_price_download
+            get :heat_network,              to: 'curves#heat_network_profiles'      # backwards compatible
```
louispt1, line 112: This doesn't work right? Shouldn't the alias be to `district_heating_profiles` @noracato @aaccensi?

noracato: Yep

PR #1769, opened 2026-06-18
Add an emissions 1990 query validation spec 

louispt1:
#### Context

There was a desire within the direct emissions enhancements to check if the queried total emissions equals the sum of the emissions in the emissions.csv with separate sums for co2, other_ghg and total_ghg for 1990.

#### Implemented changes

Added another spec to lib/graph_data_validation/spec/validation/emissions.rb to validate the totals queries align with actual csv data to ensure no queries are missed when adding aggregates

@noracato I'm not sure how much to abstract here? The spec looks a little over the top, but it also feels extra to make another class to hold the logic. If you think that's the right approach I'm happy to extract an emissions_csv_reconciler class in the lib to clean up the spec.

## Checklist

- [X] I have tested these changes
- [X] I have tagged the relevant people for review

### DIFF
--- a/lib/graph_data_validation/spec/validation/emissions.rb
+++ b/lib/graph_data_validation/spec/validation/emissions.rb
@@ -18,5 +18,41 @@ RSpec.shared_examples "emissions" do |dataset|
       end
     end
   end
-end
 
+  # Reconcile the 1990 direct_emissions total queries against the dataset's
+  # emissions.csv (read from the local ETSource).
+  #
+  # Bunkers is excluded on both sides: its query scales the inventory by the
+  # bunkers_allocated_percentage_* inputs (default 0), so it cannot equal the raw
+  # CSV. Subtracting the bunkers query and the Bunkers CSV rows makes the check
+  # allocation-independent.
+  #
+  # LULUCF removals are stored as positive values in the CSV but represent a
+  # carbon sink; the queries subtract them (NEG), so the CSV sum negates them too.
+  emissions = Atlas::Dataset.find(dataset.scenario.area_code).emissions
+
+  [
+    { total: 'direct_emissions_co2_1990',       bunkers: 'direct_emissions_bunkers_co2_1990',       ghgs: %w[co2] },
+    { total: 'direct_emissions_other_ghg_1990', bunkers: 'direct_emissions_bunkers_other_ghg_1990', ghgs: %w[other_ghg] },
+    { total: 'direct_emissions_total_ghg_1990', bunkers: 'direct_emissions_bunkers_total_ghg_1990', ghgs: %w[co2 other_ghg] }
+  ].each do |check|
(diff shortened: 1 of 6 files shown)


kndehaan:
The functionality of the test works well! I have one suggestion regarding the nr of validated datasets.

The purpose of the test is to check whether all EMISSIONS() rows are included in the co2, other_ghg and total_ghg queries. It is currently tested for all country datasets, but just checking a few datasets would already suffice and will be faster. 

We could define a couple of datasets in `config.yml` for which the queries should be tested with the emissions.csv and only do the test for those datasets:

```
emissions_1990_validation:
- nl2023
- nl2019
- DE_germany
- UKNI01_northern_ireland
```

Would it be possible to setup the spec up in a way to only test these few datasets?

lib/graph_data_validation/spec/validation/emissions.rb
```diff
@@ -18,5 +18,41 @@
...
+      next 0.0 unless row[:year] == 1990 && check[:ghgs].include?(row[:ghg])
+      next 0.0 if row[:etm_sector].to_s.casecmp?('bunkers')
+
+      value = row[:value].to_f
+      removal = row[:etm_sector].to_s.casecmp?('lulucf') &&
+                row[:etm_subsector].to_s.casecmp?('removals')
+      removal ? -value : value
+    end / 1000.0
+
+    context "1990 inventory reconciliation for #{check[:total]}" do
```
kndehaan:
```suggestion
    context "1990 emission results for #{check[:total]}" do
```

noracato (commented):
If been looking and playing with it a bit. 

This is adding query and atlas data validation to this lib. That is why it does not fit.

This lib was quickly designed for modellers to be able to themselves add tests for nodes. I don't think it is a bad idea to abstract the complicated looking logic out of the spec to make it more readable for modellers. Try to keep modeller logic in the spec, like:
```ruby
dataset.query("present:Q(direct_emissions_total_ghg_1990)") -
   dataset.query("present:Q(direct_emissions_bunkers_total_ghg_1990)")
```

louispt1:
> If been looking and playing with it a bit.
> 
> This is adding query and atlas data validation to this lib. That is why it does not fit.
> 
> This lib was quickly designed for modellers to be able to themselves add tests for nodes. I don't think it is a bad idea to abstract the complicated looking logic out of the spec to make it more readable for modellers. Try to keep modeller logic in the spec, like:
> 
> ```ruby
> dataset.query("present:Q(direct_emissions_total_ghg_1990)") -
>    dataset.query("present:Q(direct_emissions_bunkers_total_ghg_1990)")
> ```

I had a go - let me know what you think.

louispt1:
@kndehaan shall i just close this one? It's now unnecessary right?

kndehaan:
As discussed today, we'll leave it open, just in case, until the new mapping approach has been implemented. Then we can probably close this without merging.

PR #1775, opened 2026-07-02
Add MUSE method and spec

louispt1:
#### Context
Add a MUSE method which is the equivalent of USE but for molecule nodes. 

Goes with pull requests ...

## Checklist


kndehaan (approved):
Awesome!

PR #1777, opened 2026-07-08
Add scheme-based SECTOR/MSECTOR lookups and mapped EMISSIONS form

louispt1:
#### Context
Adds the ETEngine surface for the sector mapping feature: Enabling SECTOR lookup by external classification schemes as recorded in the sector_mapping.csv config in ETSource.

#### Implemented changes
- `Etsource::Sectors`: Rails-cached mapping hash and per-graph-type `(label, use) -> node keys` index (mirrors the merit order import).
- `Qernel::Sectors`: per-graph-instance resolver — `scheme?`, `pairs(scheme, values)`, `lookup(scheme, values)` — memoized per instance, reached via a new `Graph#sector_map` accessor.
- `SECTOR`/`MSECTOR` now dispatch on arity: one argument is the legacy namespace-sector filter (byte-for-byte unchanged); two or more is `scheme, value(s)`, returning the union of matching live nodes. Works identically on the molecule graph via `MSECTOR`.
- `EMISSIONS` dispatches on whether the first argument is a known scheme: `EMISSIONS(scheme, value, ghg[, 1990])` sums the emissions store over the mapping's resolved `(label, use)` pairs (missing store keys treated as 0); all legacy `EMISSIONS` forms are unchanged.
- Unknown scheme/value raises (`Gql::UnknownSectorSchemeError` / `UnknownSectorValueError`) naming the valid options; a valid value with no labelled nodes returns `[]`.
- Spec additions in `spec/models/gql/runtime/functions/lookup_spec.rb` covering legacy back-compat, multi-value union, pair-correctness (energetic vs non-energetic), unknown scheme/value errors, blank/`-` cells, and the mapped EMISSIONS form.

#### Related
[atlas](https://github.com/quintel/atlas/pull/198)
[etengine](https://github.com/quintel/etengine/pull/1777)
[etsource](https://github.com/quintel/etsource/pull/3459)

## Checklist

- [ ] I have tagged the relevant people for review

### DIFF
--- a/app/models/etsource/sectors.rb
+++ b/app/models/etsource/sectors.rb
@@ -4,9 +4,6 @@ module Etsource
   # Reads the sector mapping and node labels from Atlas and caches the two
   # structures GQL needs to resolve `SECTOR(scheme, value)` style queries.
   #
-  # Mirrors the merit order import precedent: the expensive Atlas read happens
-  # once and the result is memoized in the Rails cache, so per-request cost is a
-  # couple of hash lookups.
   #
   #   * mapping        - {scheme => {value => [[label, use], ...]}}
   #   * node_index(g)  - {[label, use] => [node_key, ...]} for one graph type,
@@ -19,7 +16,7 @@ module Etsource
 
     # Public: The inverted mapping index, keyed by scheme then normalized value.
     def mapping
-      Rails.cache.fetch('sector_mapping_hash') do
+      NastyCache.instance.fetch('sector_mapping_hash') do
         Atlas::SectorMapping.load.to_h
       end
     end
@@ -27,7 +24,7 @@ module Etsource
     # Public: The (label, use) -> node keys index for the given graph type
     # (:energy or :molecules).
     def node_index(graph_type)
-      Rails.cache.fetch("sector_node_index_#{graph_type}") do
+      NastyCache.instance.fetch("sector_node_index_#{graph_type}") do
         build_node_index(node_class_for(graph_type))
       end
     end
--- a/app/models/etsource/sectors.rb
+++ b/app/models/etsource/sectors.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module Etsource
+  # Reads the sector mapping and node labels from Atlas and caches the two
+  # structures GQL needs to resolve `SECTOR(scheme, value)` style queries.
+  #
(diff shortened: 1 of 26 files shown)


app/models/etsource/sectors.rb
```diff
@@ -0,0 +1,50 @@
...
+  #                      with `use` baked in at import so it is never consulted
+  #                      per query.
+  class Sectors
+    def initialize(etsource = Etsource::Base.instance)
+      @etsource = etsource
+    end
+
+    # Public: The inverted mapping index, keyed by scheme then normalized value.
+    def mapping
+      Rails.cache.fetch('sector_mapping_hash') do
```
noracato: Depending on how much the new sector is used, we might want to use process memory (NastyCache) vs the Rails cache

app/models/gql/query_interface/lookup.rb
```diff
@@ -64,6 +64,22 @@ def molecule_sector_nodes(keys)
         molecule_graph_helper.sector_nodes(keys)
       end
 
+      # Scheme-based sector mapping lookups. `energy`/`molecule` select the
+      # graph whose live nodes are returned; the mapping itself is shared.
+      def energy_sector_map(scheme, values)
```
noracato:
```suggestion
      def energy_sector_map_nodes(scheme, values)
```
This method is returning nodes, not a map. It's more clear to put that already in the methods name

app/models/gql/runtime/functions/lookup.rb
```diff
@@ -374,6 +397,28 @@ def EMISSIONS(*keys)
...
+        end
+
+        suffix    = ('1990' if year.to_i == 1990)
+        emissions = scope.graph.emissions
+
+        scope.sector_resolver.pairs(scheme, value).sum do |label, use|
+          key = [label, use, ghg, suffix].compact.join('_').to_sym
+          emissions[key] || 0.0
+        end
+      end
```
noracato: Let's think about if summing logic should live in GQL lookup. Did you consider a (class) method on emissions?

noracato (commented):
Nice work! Already left a few questions for you to think about. 

I was also contemplating if the `Qernel::Sector` could benefit from `Dry`, what do you think about that?

kndehaan:
Good for official review, rerequesting @noracato 's review.

PR #1779, opened 2026-07-14
Emissions

louispt1:
#### Context
Adds multi-year emissions support with 1990 baseline tracking and aggregation capabilities.

#### Implemented changes
  - Added `year` column to emissions CSV schema
  - Implemented `Qernel::Emissions#sum()` for subsector aggregation
  - Updated `EMISSIONS()` GQL function to aggregate across subsectors
  - Enhanced `ScopedSector` with year-aware scoping
  - Updated atlas gem dependency (includes emissions schema changes)
  - Add a MUSE method which is the equivalent of USE but for molecule nodes. 

#### Related
Goes with pull requests:
- https://github.com/quintel/atlas/pull/199
- https://github.com/quintel/etengine/pull/1779
- https://github.com/quintel/etsource/pull/3463
- https://github.com/quintel/etmodel/pull/4753
- https://github.com/quintel/etdataset/pull/1095
- https://github.com/quintel/documentation/pull/343

## Checklist


app/models/qernel/node_api/direct_emissions.rb
louispt1, line 1: @kndehaan Do you mind just double checking that the methods added (also the molecule emissions methods) are still required? I vaguely remember you saying they might not be needed anymore?

kndehaan: We're not using them yet in this increment but we do want to use them eventually. Therefore, it's fine to leave them in!

kndehaan (approved):

mabijkerk:
I had AI do a full review of the branches, it flagged the following:

**The CCS-capture spec never runs** — [direct_emissions_spec.rb:91](etengine/spec/models/qernel/node_api/direct_emissions_spec.rb:91)

`pending 'with CCS capture (free_co2_factor)' do ... end` defines a pending
*example*, not a group, so the `let` calls inside it raise `` `let` is not
available from within an example `` and the whole block is swallowed as
"pending, no reason given". This is the one behaviour I'd most want covered
on this branch: `direct_reporting_emissions_co2_capture` appears in every
present-year net-CO2 query (`production - capture`). It wants `context` with
a `skip`/`pending` inside, or `xcontext`, instead of `pending` wrapping the
whole block.

kndehaan:
Thanks @mabijkerk. I propose to not let this block merging the PR for now, but I'll ask @louispt1 next week why this one's not running.

louispt1:
> I had AI do a full review of the branches, it flagged the following:
> 
> **The CCS-capture spec never runs** — [direct_emissions_spec.rb:91](etengine/spec/models/qernel/node_api/direct_emissions_spec.rb:91)
> 
> `pending 'with CCS capture (free_co2_factor)' do ... end` defines a pending _example_, not a group, so the `let` calls inside it raise `` `let` is not available from within an example `` and the whole block is swallowed as "pending, no reason given". This is the one behaviour I'd most want covered on this branch: `direct_reporting_emissions_co2_capture` appears in every present-year net-CO2 query (`production - capture`). It wants `context` with a `skip`/`pending` inside, or `xcontext`, instead of `pending` wrapping the whole block.

The pending spec can indeed be removed as its no longer relevant. I did notice that the existing specs only check edge zeroed cases, so I added a calculation case as well on etengine branch `emissions` to cover a bit more with the spec.

PR #1780, opened 2026-07-15
Revert tolerance for share group balance validation

louispt1:
#### Context
Change tolerance for share group balance validation from 1.0E-12 to 10E-2 to revert to the previous behaviour of allowing a 1% tolerance for share group balance validation.

See https://github.com/quintel/pyetm/issues/197 for more context. The stricter balancing was causing problems for API users.

## Checklist

### DIFF
--- a/app/models/scenario_updater/services/validate_balance.rb
+++ b/app/models/scenario_updater/services/validate_balance.rb
@@ -6,7 +6,7 @@ class ScenarioUpdater
     class ValidateBalance
       include Dry::Monads[:result]
 
-      TOLERANCE = 1.0E-12
+      TOLERANCE = 1.0E-2
       SHARE_GROUP_TOTAL = 1.0E2
 
       def call(scenario, user_values:, balanced_values:, provided_values:, skip_validation: false)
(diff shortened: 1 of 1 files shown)


louispt1:
@noracato 
Unfortunately force_balance doesn't work --> when provided shares sum to just over 100, the balancing algorithm tries to solve this by adjusting the unset inputs (which are already at 0). force_balance only allows overwriting previously set share groups, rather than the current set of changes.

I would suggest we catch a diff of ~ 1e-6 and rescale in that case. We can handle that either in pyetm or the engine - but probably in the engine is more robust? 

One question is if we rescale with each value by 100/`share sum` (leading to lots of odd numbers with long decimal trails) or if we use an inverse of the "Hare-Niemeyer method" used in the amalgamator to redistribute the 'remainder' of a share group's total based on the 'entitlement' weights of member shares.

noracato:
THB if we really want to solve it clean, add a case to [quintel/osmosis](<https://github.com/quintel/osmosis>) for when all values are marked static with a simple 100/`share sum` rescale. Then `autobalance` will just take care of it. And if people don't want their shares autobalanced, they can already turn it off

louispt1:
See https://github.com/quintel/etengine/pull/1786

PR #1781, opened 2026-07-15
Replace ctrl+enter with cmd+enter for mac on firefox

louispt1:
#### Context
Because of Firefox issues with ctrl+enter (see: https://bugzilla.mozilla.org/show_bug.cgi?id=237027) I have extended the command to also accept cmd+enter. For me this works perfectly well and can be considered an MVP solution.

Note: ctrl+enter still acts as a right click for me. I also have that behaviour on other pages (e.g. gmail).

What do you think @mabijkerk ?

## Checklist

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -37,9 +37,9 @@ $(document).on("turbo:load", function () {
       cm.focus();
     }
 
-    // Ctrl+Enter submit on any CodeMirror instance
+    // Ctrl+Enter (Windows/Linux) or Cmd+Enter (Mac) submit on any CodeMirror instance
     cm.on("keydown", function (el, event) {
-      if (event.ctrlKey && event.keyCode === 13) {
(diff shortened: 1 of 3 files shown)


mabijkerk:
I like it and it works, thanks for fixing it.

Only question: wouldn't we rather kill the ctrl + return function then?

louispt1:
Well only if we are always only on mac as admins? Which is currently true but maybe not always?

mabijkerk:
On Windows the Cmd button is replaced by the Windows button right? As a non-confusing alternative, shouldn't we just facilitate Option (Mac) and Alt (Windows)?

louispt1:
> On Windows the Cmd button is replaced by the Windows button right? As a non-confusing alternative, shouldn't we just facilitate Option (Mac) and Alt (Windows)?

@mabijkerk I now changed it to alt/option only in the sandbox

mabijkerk (approved):
Nice work!

mabijkerk:
I replaced the symbol with text, since I thought that would be slightly more readable.

PR #1786, opened 2026-07-21
Repair share-group drift within a reasonable tolerance

### LINKED ISSUE #197: ETEngine fails to make scenarios from Excel due to floating point drift
#### Context
I am creating a scenario from the MAIN tab of an existing Excel file. Here I have explicitly summed every share sum to 100.0

Upon `pyetm run` ETEngine throws a 422 share group error because multiple share groups sum to 100,0000000001. The likely culprit (according to Claude) seems to be floating point drift (rounding errors) because of the conversion from Excel.

#### Steps to reproduce
Create a new scenario (here: for RES Drechtsteden) on production with the following parameters:
agriculture_bio_kerosene_in_crude_oil_share 0.0
agriculture_biodiesel_in_crude_oil_share		53.5
agriculture_diesel_in_crude_oil_share		43.0
agriculture_kerosene_in_crude_oil_share 0.0
agriculture_lpg_in_crude_oil_sharea		3.5
agriculture_other_bio_oil_in_crude_oil_share    0.0
agriculture_other_oil_in_crude_oil_sharea		0.0

Version (e.g. production/2025-01): production

#### Proposed solution
Perhaps some sort of share group validation step can be incorporated into `pyetm run` that catches these errors?

louispt1:
#### Context
A user who builds a scenario in Excel might see that a share group sums to exactly 100.0, but then ETEngine can reject the change with a 422:
```
"agriculture_crude_oil" group does not balance: group sums to 100.0000000001
```

The drift is introduced by floating-point arithmetic in Excel itself. [See this pyetm issue for an example](https://github.com/quintel/pyetm/issues/197).

Widening the tolerance does not address the underlying issue, nor does correcting the floating points in pyetm (as other API users could still encounter the same problems).

Solving this problem within Osmosis is also not correct, because Osmosis' concept of static and dynamic shares would be violated by redistributing the difference across the group.

#### Implemented changes
- When Osmosis reports that a group cannot balance, Balancer considers whether the deviation is small enough that nobody could have 'meant it'. If the total is within `Balancer::INTENT_TOLERANCE` of 100, every member is rescaled **value-proportionally** (`× 100/total`) in exact Rational arithmetic
-  Both `NoVariablesError` and `CannotBalanceError` trigger the re-balancing 
- The `INTENT_TOLERANCE` variable is set to 1e-6 because: the finest level of step value is 0.01. The biggest floating point drifts should be 1e-10. 1e-6 should catch drift while preserving intent.
- Every repair logs once at info with scenario, group and exact deviation.
- Spec accounts for the majority of the additions in the diff

#### Related
Fixes https://github.com/quintel/pyetm/issues/197

Goes with pull requests ...
#1780 should be closed unmerged, the solution is incorrect

## Checklist

### DIFF
--- a/app/controllers/api/v3/scenarios_controller.rb
+++ b/app/controllers/api/v3/scenarios_controller.rb
@@ -454,7 +454,7 @@ module Api
       # Returns a ActionController::Parameters
       def filtered_params
         params.permit(
-          :autobalance, :force, :reset, gqueries: []
+          :autobalance, :force, :force_balance, :reset, gqueries: []
         ).merge(scenario: scenario_params)
       end
 
--- a/app/controllers/inspect/checks_controller.rb
+++ b/app/controllers/inspect/checks_controller.rb
@@ -31,6 +31,7 @@ class Inspect::ChecksController < Inspect::BaseController
 
     # @return [true, false]
     #   Returns if the group sums up to -- or very close to -- 100.
+    #
     def ok?
       sum >= 99.9999 && sum <= 100.0001
     end
--- a/app/models/balancer.rb
+++ b/app/models/balancer.rb
@@ -1,17 +1,18 @@
 # Balances a group of inputs so that the sum of their values "balances" to a
 # chosen number (typically 100).
 #
-# Uses BigDecimal internally to prevent floating-point precision from causing
-# minor imperfections in balanced values.
+# Uses Rational arithmetic internally to prevent floating-point precision from
+# causing minor imperfections in balanced values.
 #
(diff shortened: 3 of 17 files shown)


louispt1:
The spec is currently failing due to an unrelated Inputs caching issue (inputs are not routed via NastyCache, so Input.clear! does not invalidate them). It can be "fixed" temporarily by not memoizing:  @before_inputs, @inputs_grouped, @coupling_inputs_keys and @coupling_group

I will try address this issue separately

PR #1787, opened 2026-07-21
Direct emissions 

louispt1:
#### Context
1. The direct emissions export left blank cells  for mapping categories blank - now they default to "-". (note you will need to checkout `emissions-ipcc` to see this behaviour, or update the mapping config).
2. I also noticed a bug in order_by which was ordering ipcc crt codes alphabetically, however that meant 2.10 was sorted before 2.2 and after 2.1. Now this issue is resolved.
3. A pending spec was removed and new spec was added to cover a CCS context where the plant's output carrier has explicitly zero CO2 content.

## Checklist


kndehaan (approved):
Looking good! Let's merge after deploy.

PR #1799, opened 2026-08-13
Fix for a known issue in Listen when running with docker compose

louispt1:
#### Context
1. Editing an ETSource input (min/max, query content) or gquery locally sometimes didn't show up in ETEngine/ETModel without restarting. It turns out that Docker didn't reliably see the ETSource file changes in the container. Forcing `Listen` to poll live solves that problem.
2. A related (but separate) issue was that `reload_record!` never cleared rails.cache properly, also forcing restarts. I've resolved that as part of this too.

#### Related
Goes with [this etlauncher PR](https://github.com/quintel/etlauncher/pull/10)

## Checklist

### DIFF
--- a/app/models/etsource/reloader.rb
+++ b/app/models/etsource/reloader.rb
@@ -49,8 +49,13 @@ module Etsource
 
         watched_dirs = DIRS.map(&Regexp.method(:escape)).join('|')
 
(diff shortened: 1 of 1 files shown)


aaccensi (approved):
LGTM!

PR #1801, opened 2026-08-17
Don't fail scenario migrations against the test database

louispt1:
#### Context
`db:prepare` in development also prepares the test databases, which led etlauncher to run the migrations for etengine twice when running `bin/update`. 

#### Implemented changes
1. Don't raise if no scenarios are migrated when running migrations against the test DB.
2. `SKIP_SCENARIO_CHECK=1` allows users to skip the scenarios migrated check in the case of a more empty DB, which has become much more common since we no longer use the dbbot DBs.

#### Related
Goes with [this etlauncher PR](https://github.com/quintel/etlauncher/pull/12)

## Checklist

- [ ] I have tested these changes

### DIFF
--- a/lib/etengine/scenario_migration.rb
+++ b/lib/etengine/scenario_migration.rb
@@ -5,6 +5,10 @@ module ETEngine
   module ScenarioMigration
     NoScenariosMigrated = Class.new(RuntimeError)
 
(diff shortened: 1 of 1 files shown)


aaccensi (approved):
LGTM!

PR #1808, opened 2026-09-10
Steel external coupling emissions migration

aaccensi:
#### Context

The steel external coupling node no longer derives its CO2 from the carriers it consumes; a new input sets it and the node emits nothing by default. Writes the amount the old model derived into `external_coupling_industry_metal_steel_external_coupling_node_co2_emissions` for scenarios routing steel through the node.

#### Implemented changes

The migration necessary to update all the relevant scenarios with the new input.

#### Related

This PR must be merged after the etsource PR: https://github.com/quintel/etsource/pull/3487

## Checklist


db/migrate/20260909120000_external_coupling_steel_emissions.rb
```diff
@@ -0,0 +1,172 @@
...
+      # A node emitting nothing is what the new input already defaults to, so only a positive
+      # amount is worth writing. An input rescaled for a differently-sized area may hold a NaN or
+      # an infinity, which carries through the arithmetic into a value worth writing even less.
+      emissions = emissions_of(scenario, coupling_share)
+      scenario.user_values[EMISSIONS_KEY] = emissions if emissions.positive? && emissions.finite?
+    end
+
+    # Names every area the migration passed over, and how many scenarios it left behind there.
+    print_report('Skipped scenarios of areas ETSource no longer holds:', retired_areas)
+    print_report('Skipped scenarios of areas with no dumped production:', undumped_areas)
```
louispt1, line 85: Nice touch

louispt1 (approved):
Super clear. I wonder if we might add some of these checks to the standard `etengine/lib/etengine/scenario_migration.rb` `migrate_scenarios` method to be honest, although that is separate from this effort.

What do you think about that @aaccensi

kndehaan (approved):
Migration works as it should! 

There could be some difference in scenario results before and after migration since the original setup determined CO2 in molecule graph based on recursion where loss conversions were taken into account. This original approach therefore did not represent the actual CO2 emissions that could be captured, but represented the primary CO2 emissions. 

Furthermore, if the carrier came from the external coupling transformation nodes, an emission factor of zero was assigned (debt of primary recursion in the transformation nodes). 

The migration now actually uses a more realistic approach by taking the carrier emissions instead of chain emissions. Good to note however that scenario results can deviate from what it was before.

